Issue #50, opened 2026-06-16
API foundation: zod DTOs, shared error envelope, validation plumbing


BenSeymourODB:
## What

Establish the `/api/*` conventions **once**, before the policy routes land, so
both built-in frontends and external integrators get a stable contract from day
one (`CLAUDE.md` → "the single contract for both frontends **and** for external
integrations"). Today `server/src/api/index.ts` is a one-line stub and nothing
is mounted under `/api`.

## Notes / shape

- zod request/response schemas living in `server/src/api/`; the frontend
  imports the **inferred** types (`z.infer`) — no hand-maintained interfaces,
  no runtime coupling between frontend and backend.
- A shared **error envelope** with a consistent shape and HTTP-status mapping.
  Recommend RFC 9457 `application/problem+json` or a small
  `{ error: { code, message, details } }`; zod validation issues surface as
  structured `details` rather than a stack trace.
- A Fastify validation hook (or zod-aware schema compiler) that rejects bad
  input with that envelope.
- Register the `/api` route prefix on the app and prove the wiring with a
  trivial `GET /api/meta` (or `/api/health`) under `app.inject()`. `/` and the
  static `/admin`·`/app` mounts (#40) stay owned as they are.
- No live DB required — this is the contract layer; the policy CRUD routes
  build on it.

## Acceptance criteria

- [ ] Error envelope + zod validation hook; a malformed body returns the
      envelope (not a 500/stack trace) with the right status.
- [ ] `/api` mounted; a trivial meta/health route proves it under
      `app.inject()`.
- [ ] Inferred request/response types exported for the SvelteKit frontend;
      the export path is documented.
- [ ] Coverage gate stays green.

## References

- `CLAUDE.md` → `api/` module, "Validate all external input with zod"
- `docs/architecture.md` → "External integrations" (the contract is shared)
- #5 (app factory the prefix mounts onto), #10 (settings)
- `docs/roadmap.md` → Phase 2

## Dependencies & blockers

_Phase-2 sequencing pass (2026-06-16). Contract layer — can run in parallel
with the schema/DB work since it needs no live database._

- **Blocked by:** nothing new — builds on the merged #5 (app factory) and #10
  (settings).
- **Blocks:** #51 (the policy routes reuse this envelope + DTO plumbing), #52
  (login/logout return errors through this envelope), #53 (the admin UI imports
  the inferred types).
[...]

PR #128, opened 2026-06-17
test(api): make the via-buildApp meta test hermetic (fixes red main) (#127)

### LINKED ISSUE #127: main CI red: `via buildApp` meta test opens the default /data/policy.sqlite
## What

The **Unit tests** CI job has been failing on `main` since the #50 API-foundation merge (PR #68, commit `a338453`). All other CI jobs (lint, docker build, pre-commit, shellcheck, ansible-lint) pass — only `vitest (unit, with coverage gate)` is red.

## Root cause

`server/tests/api/plugin.test.ts` → `describe("GET /api/meta (via buildApp)")` builds the app with default settings and **no injected db**:

```ts
app = buildApp({ settings: loadSettings({ PCT_LOG_LEVEL: "silent" }) });
```

With no db handle injected, `buildApp` calls `createDb(settings)` (`src/web/app.ts`), which opens `settings.databaseUrl` — defaulting to `/data/policy.sqlite`. On a CI runner (and any dev box without the mounted data volume) `/data` doesn't exist, so better-sqlite3 throws:

```
TypeError: Cannot open database because the directory does not exist
 ❯ new Database node_modules/better-sqlite3/lib/database.js:65:9
 ❯ createDb src/policy/db.ts:76
 ❯ buildApp src/web/app.ts:71
 ❯ tests/api/plugin.test.ts:132 (beforeEach)
```

The three tests in that block fail in `beforeEach`, and the `afterEach` then trips on `app.close()` of an undefined `app`.

## Fix

Pass `DATABASE_URL=:memory:` into `loadSettings` for that block so it still exercises the real `buildApp` wiring (the block deliberately doesn't inject a db handle) while staying hermetic — no dependency on the data volume existing. This is **not** a test weakening: the intent (prove `/api/meta` is mounted via the real `buildApp`) is preserved.

## Notes

- Caught by an `implement-issue` run while basing #82 on `main`; the same red job would block every subsequent PR's CI.
- Fixed in a dedicated PR off `main`; #82's branch is stacked on top of that fix so its own CI is green.

BenSeymourODB:
## Summary

- Fixes the **Unit tests** CI job that has been red on `main` since the #50 merge (#68).
- The `GET /api/meta (via buildApp)` block opened the default `/data/policy.sqlite` (no db injected → `buildApp` runs `createDb`), which fails on any box without the mounted data volume. Now it loads `DATABASE_URL=:memory:`.

## Why this isn't a test weakening

The block intentionally exercises the **real** `buildApp` wiring (rather than the `buildTestApp` helper) to prove `/api/meta` is mounted. Injecting a handle would change that intent; pointing `buildApp`'s own `createDb` at an in-memory database keeps the intent and makes it hermetic.

## License-boundary note

N/A — test-only change; no transport, packaging, or dependency changes.

## Test plan

- [x] `npm test` passes with **no** `/data` directory present (190 tests, 100% lines / 95% branch).
- [x] `format:check`, `lint`, `typecheck` green.

Closes #127

---
_Generated by [Claude Code](https://claude.ai/code/session_01TyTvMhj5pjLsxmXp7ek242)_

### DIFF
--- a/server/tests/api/plugin.test.ts
+++ b/server/tests/api/plugin.test.ts
@@ -129,7 +129,13 @@ describe("GET /api/meta (via buildApp)", () => {
   let app: FastifyInstance;
 
   beforeEach(() => {
-    app = buildApp({ settings: loadSettings({ PCT_LOG_LEVEL: "silent" }) });
+    // In-memory DB so buildApp's own createDb (we deliberately don't inject a
+    // handle here — the point is to exercise the real buildApp wiring) doesn't
+    // try to open the default /data/policy.sqlite, which doesn't exist on a CI
+    // runner or a dev box without the mounted data volume.
+    app = buildApp({
+      settings: loadSettings({ PCT_LOG_LEVEL: "silent", DATABASE_URL: ":memory:" }),
+    });
   });
 
   afterEach(async () => {


BenSeymourODB:
Closing as redundant - see commit from #102

Issue #133, opened 2026-06-17
Collision-resistant drizzle migrations (timestamp prefix) for concurrent sessions


BenSeymourODB:
## Problem

drizzle-kit's default `index` prefix numbers migrations sequentially
(`0000_`, `0001_`, `0002_…`). When two sessions branch off the same `main`
and each generates a migration, **both pick the same number** — and not just
the same number, the same **filenames**: `000N_<name>.sql` *and*
`meta/000N_snapshot.json`. Whoever merges second hits an add/add conflict on
the SQL, the snapshot, and `_journal.json`, and has to manually renumber +
regenerate.

This is not hypothetical — it bit PR #123 **twice in a single session**: the
schedules `ordinal` migration had to be renumbered `0001 → 0002` first against
the auth merge (#102) and again after re-merging `main`. The sibling project
hit the same class of pain and solved it: see
[next-digital-wall-calendar#346](https://github.com/BenSeymourODB/next-digital-wall-calendar/issues/346)
(switched Prisma to datetime-prefixed migration names + a naming-check script
+ a drift CI). The same coordination tax shows up in the repeated, independent
"make the meta test hermetic" fixes (#127 / PR #128 vs the fix already on
`main`).

## Proposed fix (drizzle translation of the #346 approach)

Set a timestamp prefix in `server/drizzle.config.ts`:

```ts
export default defineConfig({
  // …
  migrations: { prefix: "timestamp" },
});
```

Verified locally on the pinned `drizzle-kit@0.31.5`: `db:generate` then emits

```
drizzle/20260617040124_slow_devos.sql
drizzle/meta/20260617040124_snapshot.json
```

Because **both** the SQL and the snapshot are timestamp-named, two concurrent
branches produce **non-colliding files** — the painful add/add overwrite
disappears entirely. The only residual merge is appending each branch's entry
to `_journal.json` (and the `idx` counter), which is a trivial, non-destructive
textual merge, not a file clobber.

Existing `0000_`/`0001_`/`0002_` migrations stay as-is — drizzle orders by the
journal's `idx`/`when`, and mixed prefixes coexist; only **new** migrations get
timestamps.

## Scope / acceptance criteria

- [ ] `migrations: { prefix: "timestamp" }` in `drizzle.config.ts`; a one-line
      comment pointing at this issue.
- [ ] CI naming check (mirroring #346): validate new migrations match
[...]

Issue #76, opened 2026-06-16
client/install-client.sh — one-command Linux Mint enrolment (orchestrator)


BenSeymourODB:
## What

Implement the top-level `client/install-client.sh` per [`docs/client-install.md`](../blob/main/docs/client-install.md) — the single command that turns a fresh **Linux Mint (Cinnamon)** desktop into an enrolled, supervised client. This is the orchestrator that sequences the other Phase 3 deliverables; it should be idempotent and re-runnable.

## Scope

- Argument/usage handling: dashboard URL + enrolment token (consumes the endpoint in the enrolment-token issue), target supervised Linux user(s).
- Pre-flight: confirm Mint/Ubuntu-family, root/sudo, network reachability to the dashboard.
- Sequence the sub-steps (each its own issue): provision the `pct-agent` service user + sudoers, install/baseline-configure the upstream tools, register the client with the dashboard, then run the self-test.
- Clear, non-punitive console output consistent with the household framing in [`README.md`](../blob/main/README.md) and `docs/client-install.md`.

## Out of scope

- Any tamper-resistance / lockdown work beyond the documented bounded posture in `docs/client-install.md` ("Tamper resistance is deliberately bounded"). Do **not** add hardening beyond that ceiling.
- The `pct-client-bridge` / `pct-client-agent` `.deb` (Phase 8b) — this script only lays down the enforcement/telemetry tooling; the agent install can be a later hook.

## Notes

- This is the Phase 3 anchor in [`docs/roadmap.md`](../blob/main/docs/roadmap.md). Depends on the enrolment-token endpoint; the install steps that follow are tracked separately so each is a single PR's worth of work.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #39, opened 2026-06-16
Container first-run setup in the entrypoint (schema migration, Ansible venv, AdGuard fetch, SSH keygen)


BenSeymourODB:
## What

`server/docker-entrypoint.sh` (added in #6) currently performs only the
in-scope first-run work: it reconciles the documented `/data` volume layout
and then `exec`s the compiled server. The remaining first-run steps that
`docs/server-deployment.md` → "First-run setup" enumerates were deferred from
#6 because each depends on a transport/feature that lands in a later roadmap
phase. This issue tracks wiring them into the entrypoint as those phases land.

## Deferred steps (each lands with its phase)

- **Schema migration** — apply the committed drizzle-kit migrations against
  `/data/policy.sqlite` on start. Lands with the runtime DB connection in
  **Phase 2** (apply via drizzle-orm's `better-sqlite3` migrator at runtime
  rather than shipping the `drizzle-kit` dev dependency in the image; see also
  the `DATABASE_URL` contract in #34). **Owned by #49** — coordinate so the
  entrypoint and the in-process migrator don't double-migrate.
- **Ansible venv bootstrap** — if `/data/ansible/venv` is missing, create it
  and `pip install ansible-core` from PyPI at runtime; sync `playbooks/` from
  the image. **Phase 6.**
- **AdGuard Home bootstrap** — driven by `PCT_ADGUARD_MODE`: `managed` fetches
  + checksum-verifies the upstream release into `/data/adguard/` and
  supervises it; `external` validates REST reachability; `disabled` does
  nothing. **Phase 7.**
- **SSH key bootstrap** — generate `/data/secrets/ssh/id_ed25519` if absent.
  Pairs with the SSH transport. **Phase 4.**

## Acceptance criteria

- [ ] Each step above implemented idempotently in the entrypoint (or a script
      it calls), guarded so a missing optional download (no network) leaves the
      dashboard starting with the affected feature disabled and an error
      surfaced in the admin UI, per `docs/server-deployment.md`.
- [ ] No GPL binary is added to the image (Ansible stays in the `/data` venv;
      AdGuard Home is fetched to `/data`, never bundled) — `license-guard`
      stays green.
- [ ] Tests for the idempotency/branching logic.

## References

- #6 (introduced the entrypoint skeleton), #34 (`DATABASE_URL` contract)
- `docs/server-deployment.md` → "First-run setup", "Volume layout"
- `docs/licensing-analysis.md`

[...]

Issue #119, opened 2026-06-16
Reverse-proxy + TLS deployment guide for non-LAN deployments


BenSeymourODB:
## What

Document (and provide example config for) running the dashboard behind a reverse proxy with TLS, for deployments beyond a trusted LAN. Per Phase 11 in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).

## Scope

- Guidance + example config (e.g. Caddy/nginx/Traefik) terminating TLS in front of the Fastify container.
- Cover the WebSocket event stream (`/api/events/stream`) proxying, and the `/api`, `/admin`, `/app` paths.
- Note interaction with the home-NAT connectivity model (clients connect **out** to the server — see [`docs/architecture.md`](../blob/main/docs/architecture.md)).

## Notes

- Phase 11 "Reverse-proxy + TLS instructions for non-LAN deployments" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Extends [`docs/server-deployment.md`](../blob/main/docs/server-deployment.md).

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Picked after a triage pass found the phase-2/3 backend tickets either covered by open PRs, blocked by the admin-UI work, or just claimed by concurrent runs, and the code-review refactor tickets (#224/#225/#226) currently conflict with in-flight PRs (#203/#213/#217) that edit the same files. #119 is a self-contained Phase-11 documentation deliverable with no open-PR conflict.

Plan:
- New `docs/reverse-proxy-tls.md`: a TLS-terminating reverse-proxy guide for non-LAN access, with copy-pasteable Caddy, nginx, and Traefik configs that proxy `/`, `/admin`, `/app`, `/api`, and crucially the `/api/events/stream` **WebSocket** (Upgrade/Connection headers).
- Make the home-NAT boundary explicit: the proxy fronts only the dashboard's HTTP surface (client→server); the SSH transport (server→client) stays on the LAN and is unaffected.
- Document the accurate runtime caveats grounded in the code: the session cookie is `HttpOnly; SameSite=Strict` and not `Secure` (TLS is enforced at the proxy), and the per-IP failed-attempt limiter keys on `request.ip` with no `trustProxy` configured — so behind a proxy it collapses to a global limiter unless real client IPs are forwarded. Will file a follow-up issue for an opt-in `trustProxy` setting.
[...]

PR #153, opened 2026-06-18
feat(api): client enrolment-token endpoint + registration (#77)

### LINKED ISSUE #77: Enrolment-token endpoint + client registration on the dashboard
## What

Add the server side of client enrolment: an admin-issued, short-lived **enrolment token** and the `/api/*` endpoint the install script calls to register a new `Client` and exchange the token for the long-lived SSH/event credentials.

## Scope

- Admin action (and `/api/*` route) to mint a single-use, expiring enrolment token.
- `POST /api/clients/enroll` consuming the token: creates the `Client` row (`hostname`, `ssh_user`, `enrolled_at`), records the supervised `UserOnClient` mappings (`linux_username`, `linux_uid`), and returns what the client needs (e.g. the server's SSH public key to authorize, the per-client bearer token for the future event stream).
- zod request/response DTOs in `server/src/api/`, behind the admin auth guard (#52) for token issuance.

## Notes

- Implements the "Enrolment-token endpoint on the dashboard" bullet of Phase 3 in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Consumed by `client/install-client.sh` (#76).
- `Client` / `UserOnClient` schema comes from the Phase 2 policy store (#51); this issue adds the enrolment flow on top.
- The per-client bearer token issued here is what the Phase 8b event stream (`/api/events/stream`) authenticates against — see [`docs/architecture.md`](../blob/main/docs/architecture.md) transport table.

BenSeymourODB:
## Summary

Adds the dashboard half of client enrolment (Phase 3):

- **`POST /api/clients/enrolment-tokens`** — **admin-guarded**. Mints a single-use, short-TTL token bound to the policy-user ↔ Linux-account mapping the admin is provisioning. Stores only the SHA-256 hash; returns the plaintext **once**.
- **`POST /api/clients/enrol`** — **token-authenticated, not session-guarded** (the install script #76 has no admin session). Validates the `Authorization: Bearer <enrolment-token>` by hash lookup, then in **one transaction** creates the `Client` + `UserOnClient` links, issues a per-client bearer token (the future `/api/events/stream` credential), and consumes the token. Returns the bearer token (once) and the server SSH public key when available.

## Plan

Linked plan: `.claude/plans/issue-77-enrolment-token-endpoint.md`
Roadmap: `docs/roadmap.md` → Phase 3 ("Enrolment-token endpoint on the dashboard"); contract in `docs/client-install.md` → step 7–8.

## Design notes

- **Schema:** new `enrolment_tokens` table + nullable `clients.bearer_token_hash` (timestamp-prefixed migration via `db:generate`, #133; drift check green). Only hashes are stored — never plaintext (mirrors `integration_tokens`).
- **Secrets:** enrolment + per-client bearer tokens are 256-bit random values, hashed with **SHA-256** (`auth/secret-token.ts`) — the right primitive for high-entropy secrets, deliberately distinct from Argon2id (which exists to slow brute force against low-entropy passwords).
- **SSH public key:** new `PCT_SSH_PUBLIC_KEY_PATH` setting. The key pair itself is generated server-side as a **Phase-4** first-run step (#39); until then the file is absent and the enrol response degrades to `sshPublicKey: null` rather than failing.
- **Endpoint spelling:** `/enrol` (British), matching the authoritative `docs/client-install.md` step 8, not the issue body's `/enroll`.
- **Module layout** mirrors `api/policy/` + `policy/repository.ts`: `api/clients/{dtos,service,ssh-identity,routes,index}`, data access in `policy/enrolment.ts`, crypto leaf in `auth/secret-token.ts`.

## Security note

`POST /api/clients/enrol` is the one intentional **unauthenticated-by-session** write. It is gated by a 256-bit, hashed-at-rest, single-use, expiring token; the supervised-user set must exactly match what the token was minted for, and each bound user is re-checked to still exist. Per-IP rate-limiting on enrol is a possible follow-up (token entropy makes brute force infeasible) — noted, not built.

## License-boundary note

No GPL code imported and no GPL binary added. No transport calls — SSH keygen is Phase 4; this PR only *reads* a public-key file if present. `license-guard` unaffected.

## Test plan

- [x] `format:check`, `lint`, `typecheck`, `db:check` (drift) all clean
- [x] `npm test` — **389 passing**; coverage 99.74% stmts / 96.85% branch (gate 80%). New `api/clients/*`, `auth/secret-token.ts`, `policy/enrolment.ts` covered.
- [x] `secret-token` (mint/hash/compare), `ssh-identity` (present/absent/empty/unreadable), `policy/enrolment` (token lifecycle + **transaction rollback** on a duplicate UID), `schema`/`migrations` (new table + column, ON DELETE set null)
- [x] HTTP matrix via `app.inject()`: mint anon→401, bad userId→404, dup username→400, happy→201 (hash-only); enrol missing/invalid/expired/used bearer→401, user mismatch→400, deleted-user→409, duplicate hostname→409, happy→201 (+links, +bearer, +sshPublicKey null/set/unreadable)

## Deferred (tracked)

- Wiring into the orchestrator `install-client.sh` → #76 (this unblocks it).
- Server-side SSH keypair generation → Phase 4 (#39 entrypoint step).
- Per-IP rate-limit on enrol → follow-up note above.

Closes #77

---
_Generated by [Claude Code](https://claude.ai/code/session_01K4LjzUBhkoQJiJUJ9YfbBZ)_

### DIFF
--- a/.claude/plans/issue-77-enrolment-token-endpoint.md
+++ b/.claude/plans/issue-77-enrolment-token-endpoint.md
@@ -1,0 +1,100 @@
+# Issue #77 — Enrolment-token endpoint + client registration
+
+Phase 3 (server side). Adds the dashboard half of client enrolment: an
+admin-minted, single-use, short-TTL enrolment token, and the endpoint the
+install script (#76) calls to register a `Client` + its `UserOnClient`
+mappings and receive what it needs (server SSH public key, per-client bearer
+token).
+
+Authoritative spec: `docs/architecture.md` → "Policy model"/transport table,
+`docs/client-install.md` → step 7–8. Builds on #50 (DTO/envelope), #49 (`app.db`),
+#52 (`requireAdmin`), #48/#51 (`clients`, `users_on_clients`).
+
+## Endpoints
+
+1. **`POST /api/clients/enrolment-tokens`** — **admin-guarded** (`requireAdmin`).
+   Mints a single-use token bound to the supervised-user mapping the admin is
+   provisioning.
+   - Body: `{ supervisedUsers: [{ userId, linuxUsername }], ttlSeconds?, hostname? }`.
+   - Validates every `userId` exists → `404` if any missing.
+   - Mints a 256-bit token (`crypto.randomBytes(32).base64url`), stores only its
+     SHA-256 hash + the mapping payload (JSON) + `expiresAt`; returns the
+     plaintext **once** (`201`).
+
+2. **`POST /api/clients/enrol`** — **token-authenticated** (NOT admin; the client
+   has no session). `Authorization: Bearer <enrolment-token>`.
+   - Body: `{ hostname, sshUser, supervisedUsers: [{ linuxUsername, linuxUid }] }`.
+   - Validates the bearer token: hash lookup → missing/expired/consumed all →
+     `401` with distinct codes. The linuxUsername set must equal the minted
+     mapping's set → `400 enrolment_user_mismatch` otherwise. Re-checks each
+     minted `userId` still exists → `409` if deleted since mint.
+   - In **one transaction**: create the `Client`, insert the `UserOnClient`
+     rows (joining minted `userId` ↔ request `linuxUid` on `linuxUsername`),
+     mint+store a per-client bearer-token hash, mark the enrolment token
+     consumed. Duplicate hostname → `409` (already enrolled).
+   - Returns `201 { clientId, hostname, sshUser, bearerToken (once),
+     sshPublicKey: string | null, supervisedUsers }`. `sshPublicKey` is read
+     from `PCT_SSH_PUBLIC_KEY_PATH` if present, else `null` (Phase-4 keygen
+     may not have run yet — degrade gracefully).
+
+Spelling: use **`/enrol`** (British), matching `docs/client-install.md` step 8
+(authoritative), not the issue body's `/enroll`. Noted in the PR.
+
+## Schema + migration
+
+- New `enrolment_tokens` table: `id`, `token_hash` (unique), `hostname`
+  (nullable), `supervised_users` (JSON), `expires_at`, `created_at`,
+  `consumed_at` (nullable), `consumed_client_id` (nullable FK → clients,
+  `set null`).
+- `clients`: add nullable `bearer_token_hash` (only enrolled clients have one;
+  CRUD-created clients (#51) legitimately don't).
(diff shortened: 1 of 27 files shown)


BenSeymourODB:
### First-pass review + resolutions

Ran a deep, security-focused self-review (the enrol endpoint is intentionally unauthenticated-by-session). **No blockers.** Addressed in `785ad26`:

- **Unique index on `clients.bearer_token_hash`** — makes the future `/api/events/stream` lookup-by-hash single-row by construction (NULLs stay distinct, so admin-CRUD clients are unaffected). Regenerated as one clean timestamp-prefixed migration; drift-clean.
- **Audit logging on the unauthenticated surface** — `info` on a successful mint/enrol, `warn` on every token rejection (invalid/used/expired), so abuse is observable.
- **Defence-in-depth single-use guard** — the consume `UPDATE` now matches `consumed_at IS NULL` and asserts one row, rolling the whole transaction back (`EnrolmentTokenConsumedError`) if a token's state changed under us. Not reachable on today's synchronous request path (the service does no `await` between check and consume), but it makes a future async refactor safe-by-construction.
- **Verified safe as-is:** token entropy (256-bit) + hash-only-at-rest + plaintext-never-logged; the supervised-user **bijection** (userId is always taken from the *minted* side — a client can't remap to an arbitrary policy user, nor drop/smuggle a user); atomic consume+enrol; no info leakage in errors.

**Deliberately deferred (documented in `routes.ts`, tracked):**
- App-layer **rate limiting** on `/clients/enrol`. The 256-bit token makes online guessing infeasible; volumetric abuse of the unauthenticated surface belongs at the reverse proxy (Phase 11). Rejections are logged so it's observable. Follow-up to be tracked as its own issue.

**Nits:** renamed the service's `EnrolResult` → `EnrolServiceResult` (disambiguate from the repo type), normalised "enrollment" → "enrolment" in the docs, noted the second-granularity expiry.

**Added tests** for the highest-value gaps the review flagged: multi-user bijection (drop → 400, smuggle → 400, both → 201 with each minted `userId` mapped to its uid), empty-list → 400, TTL max-reject + default-applied, and the repo-level consume-guard rollback. **396 tests pass; coverage 99.5% stmts / 96.7% branch.** All CI checks were green on the prior push.

---

BenSeymourODB:
Deferred rate-limiting follow-up is now tracked in #154 (and the reverse-proxy/DoS story is #119).

---

Issue #88, opened 2026-06-16
Normalise AW events into UsageSample + budget rollup/aggregation views


BenSeymourODB:
## What

Turn pulled ActivityWatch events into `UsageSample` rows mapped to the policy's `Activity` matchers, and build the aggregation views the dashboard renders. This is the data half that the Phase 5 chart components (#62) and the "My Time" client dashboard (#61) consume.

## Scope

- Map raw AW window/afk events → `UsageSample` (`user_id`, `client_id`, `activity_id`, `started_at`, `ended_at`), resolving the app/domain against `Activity`/`ActivityGroup` matchers. Discard raw blobs once normalised.
- Rollup/aggregation queries for the derived views in [`docs/architecture.md`](../blob/main/docs/architecture.md): per-user budget **burndown** (today/week/month) and per-activity **timeline**.
- Budget windows computed in the user's **effective timezone**, pinned per the mid-window rule — reuse `server/src/policy/budget-window.ts` ([ADR 0001](../blob/main/docs/adr/0001-budget-timezone.md), [ADR 0003](../blob/main/docs/adr/0003-mid-window-timezone-change.md)).
- Conservative on gaps: missing telemetry credits **no** consumption, never a punitive deduction.

## Notes

- Phase 5 "Normalisation into `UsageSample` rows; aggregation views" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- These rollups are also what the Phase 8 per-activity enforcement decision logic reads.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Scope for this run: the **data layer** — (1) a `UsageSample` table + Drizzle migration; (2) a pure normaliser mapping the merged AW client's (`#87`) `AwWindowEvent[]`/`AwAfkEvent[]` → `UsageSample` rows, resolving app/title against `Activity`/`ActivityGroup` matchers, with overlap dedup, clock-skew drop, and AFK-aware crediting (missing telemetry credits **no** consumption, never punitive); (3) rollup/aggregation queries — per-user budget burndown (today/week/month) and per-activity timeline — with budget windows computed in the user's effective timezone via `policy/budget-window.ts` (ADR 0001/0003).

Wiring the normaliser into the scheduled telemetry pull is deliberately **out of scope here** — that's the `consume()` seam PR #162 (telemetry port-forward) explicitly left open; it composes once both land. Admin chart rendering is #62.

---
[...]

Issue #87, opened 2026-06-16
ActivityWatch REST client + zod-validated event parsing


BenSeymourODB:
## What

The `transport/activitywatch/` REST client that, over the SSH-forwarded port (#86), queries `aw-server`'s API for events in the polling window and validates them with zod before they cross into typed code. REST-only integration — no source-level coupling to AW (license boundary, [`docs/licensing-analysis.md`](../blob/main/docs/licensing-analysis.md)).

## Scope

- Typed client for the AW buckets/events endpoints needed for window/afk activity.
- zod schemas for AW responses; reject/skip malformed payloads rather than trusting them.
- Windowed querying (pull only since the last successful sample to keep payloads bounded).

## Notes

- Phase 5 in [`docs/roadmap.md`](../blob/main/docs/roadmap.md); transport row in [`docs/architecture.md`](../blob/main/docs/architecture.md).
- Output feeds the `UsageSample` normalisation issue; raw event blobs are discarded after aggregation.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Picking up #87 (the `transport/activitywatch/` REST client + zod-validated event parsing) as the lowest-phase issue that is genuinely unblocked with a clean end-to-end TypeScript/Vitest test story. The lower-phase candidates are not eligible right now: Phase 2's #51/#53/#54 sit behind the open auth PR #102 (#52), and #59/#63 are deferred until #53 lands; Phase 3's server issues (#77/#81) need the unbuilt CRUD/auth and the rest of Phase 3 is client-side bash without a test harness in this repo; #82 (the SSH facade) was already claimed <1h ago, so its Phase-4 dependents (#83/#84/#85) aren't ready.

Scope per the issue: a typed REST client for the `aw-server` buckets/events endpoints needed for window/afk activity, zod schemas that reject/skip malformed payloads, and windowed querying (pull only since the last successful sample). The client takes an injected base URL/host:port so it is fully unit-testable against a mock server and composes cleanly with the SSH-forwarded port from #86 later. Normalisation into `UsageSample` rows stays out of scope (that's the downstream #88); raw event blobs are discarded after the caller aggregates.

---
_Generated by Claude Code_

---
_Generated by [Claude Code](https://claude.ai/code)_

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

[...]

Issue #84, opened 2026-06-16
Offline-queue: persist policy pushes for unreachable clients + replay on reconnect


BenSeymourODB:
## What

When a client is offline at policy-change time, persist the intended change and replay it on the next successful SSH probe, per the "Client offline at policy-change time" failure mode in [`docs/architecture.md`](../blob/main/docs/architecture.md).

## Scope

- A durable queue (SQLite) of pending per-client transport actions, with ordering and dedup/coalescing (a newer push for the same target supersedes an older queued one).
- Reconnect detection: a croner SSH-probe that drains the queue when a client becomes reachable.
- Idempotent replay — re-applying a drained action must be safe.
- Expose queued state so the admin Clients page (#81) and the Phase 4 "preview diff" (#64) can show what's pending.

## Notes

- Phase 4 "Offline-queue" bullet in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Conservative semantics: a missed push is queued, never silently dropped.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Selection trace: Phase 1 #39 is a cross-phase tracker (its steps land with Phases 4/6/7), not a single end-to-end deliverable. Phase 2 (#51/#148/#124/#63/#53) is entangled with the unmerged PR #156 that reshapes the `schedules`/`exceptions` tables. Phase 4 #83 is in unmerged PR #155; #64/#140/#143 depend on #146 (PR #156). That leaves #84 and #85 as the independent Phase-4 candidates; #84 is the lower number and is explicitly anticipated by the merged SSH facade (`SshUnreachableError` "is the offline-queue's signal (#84)"). It touches none of the tables/wrappers under the open PRs and is backend-only (its UI consumers #81/#64 are future), so it can land end-to-end now.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #83, opened 2026-06-16
timekpra invocations: daily/weekly/monthly limits, allowed hours, PlayTime


BenSeymourODB:
## What

The typed wrapper over the `timekpra` CLI (invoked via the SSH facade) that translates policy into Timekpr-nExT configuration: set daily/weekly/monthly session limits, set allowed hours, and configure PlayTime (app-group time). Parse `timekpra` stdout to confirm application.

## Scope

- Map `Budget` (scope `overall`, windows `daily|weekly|monthly`) → `timekpra` limit commands for the supervised `linux_username`.
- Allowed-hours / schedule windows → `timekpra` allowed-hours config.
- App-group budgets → Timekpr-nExT **PlayTime** configuration.
- zod-parse `timekpra` output; surface failures through the facade's error taxonomy.
- Pure CLI + stdout; **no** parsing of Timekpr's on-disk state with its own code (license boundary, [`docs/licensing-analysis.md`](../blob/main/docs/licensing-analysis.md)).

## Notes

- Phase 4 "`timekpra` invocations for: set daily/weekly/monthly limits, set allowed hours, set PlayTime configuration" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Enforcement responsibility mapping is in [`docs/architecture.md`](../blob/main/docs/architecture.md) "Enforcement responsibilities".
- Budget-window rollover must respect the user's effective timezone — see [ADR 0001](../blob/main/docs/adr/0001-budget-timezone.md) and the budget-window helper.

BenSeymourODB:
🤖 implement-issue claiming this for the current session.

**Why #83 this run (selection trace):** Phase 1 #39 is a cross-phase tracker (steps land with Phases 4/6/7), not end-to-end implementable now. Phase 2 is currently frozen for this run: #146 (the foundational `schedules`/`exceptions` column reservation per ADR 0005) was claimed <6h ago by a concurrent session, and #51/#148/#124 all build on or concurrently mutate those same tables — they'd collide with that in-flight migration. Phase 3 #76 waits on the unmerged enrolment PR #153, #80 runs inside the not-yet-built install script, and #81 needs transport/telemetry not yet built. #83 is unclaimed, unblocked (SSH facade #82 is merged), and sits at the transport/command layer *below* the schedule schema being reshaped, so it's independent.

**Scope for this run** (matches the issue + keeps the license boundary intact):
[...]

PR #162, opened 2026-06-18
feat(transport): scheduled SSH port-forward telemetry window to aw-server :5600 (#86)

### LINKED ISSUE #86: Telemetry pull: scheduled SSH port-forward to aw-server :5600
## What

The croner-scheduled job that, per enrolled client, opens an SSH port-forward to the client's `aw-server` on `localhost:5600` and makes it reachable to the dashboard for the polling window — the inbound half of the data flow in [`docs/architecture.md`](../blob/main/docs/architecture.md) "Inbound (client → server) — telemetry pull".

## Scope

- croner job iterating reachable clients, opening a forwarded local port via the Phase 4 `ssh2` facade.
- Robust open/close lifecycle (don't leak forwards; bounded concurrency across many clients).
- Skip/queue cleanly when a client is offline (telemetry gaps are acceptable and non-punitive per the failure-modes section).

## Notes

- Phase 5 "Scheduled job (croner) that opens SSH port-forwards and pulls AW events" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- ActivityWatch's REST API is unauthenticated and **must never be network-exposed** — all access is via this server-initiated SSH tunnel.
- Pairs with the AW REST client + normalisation issues; this one only establishes the transport window.

BenSeymourODB:
## Summary

The inbound **transport window** for Phase-5 telemetry, exactly as #86 frames it ("this one only establishes the transport window"):

- **`SshTransport.withPortForward(target, remote, fn)`** (`transport/ssh/facade.ts`) — a loopback-bound (`127.0.0.1`) ephemeral TCP listener whose connections are forwarded over the **pooled** SSH session to a remote endpoint on the client (`ssh2` `forwardOut`). The listener and any in-flight sockets are always torn down when `fn` settles (no leak past the window); a single failed forwarded connection is dropped without sinking the window; a connect failure surfaces as `SshUnreachableError` (the offline-queue signal, #84).
- **`runTelemetryPull(deps)`** (`transport/activitywatch/telemetry.ts`) — a bounded-concurrency pass over enrolled clients that opens a forward per reachable client, skips offline clients non-punitively (telemetry gap, not failure), isolates a per-client error from the rest of the pass, and returns a `{ attempted, succeeded, skippedOffline, failed }` summary. The default consumer `probeAwServer` confirms the tunnelled `aw-server` via the merged AW REST client (#87); a `consume(baseUrl, client)` seam is left for the #88 normalisation.
- **`scheduleTelemetryPull(run, opts)`** — croner wrapper with overlap protection (`protect`) and a `catch` handler, returning the `Cron` handle for shutdown. **`PCT_TELEMETRY_PULL_CRON` / `PCT_TELEMETRY_PULL_CONCURRENCY`** config, the cron pattern validated at startup (`isValidCronPattern`, mirroring `isValidTimeZone`).

## Plan

Linked plan: `.claude/plans/issue-86-telemetry-port-forward.md`
Roadmap: `docs/roadmap.md` → Phase 5 ("Scheduled job (croner) that opens SSH port-forwards and pulls AW events").

## License-boundary note

ActivityWatch is reached only over its REST API (#87), through a server-initiated SSH tunnel **bound to loopback** — AW's API is unauthenticated and must never be network-exposed. The tunnel is pure TCP forwarding via `ssh2`; no ActivityWatch/GPL source is linked in-process and no GPL binary is added to the image. **No new dependency** (`ssh2`, `croner`, `zod` already present); `license-guard` unaffected.

## Test plan

- [x] `withPortForward` — loopback round-trip over a `PassThrough` channel, remote-host default/override, dropped-forward isolation, callback-rejection propagation + teardown, `SshUnreachableError` (callback never runs), pooled-connection reuse with `exec`.
- [x] `runTelemetryPull` — success summary, configured aw-port + loopback base URL, offline-skip, consumer-error isolation, concurrency bound, empty list; default consumer `probeAwServer` against a real loopback HTTP server (healthy + malformed).
- [x] `scheduleTelemetryPull` — next-run scheduled, `trigger()` runs the pass, thrown pass routes to `catch`; `isValidCronPattern` accept/reject.
- [x] config — telemetry defaults, explicit cron + concurrency, invalid cron, non-positive/non-numeric concurrency.
- [x] `format:check` / `lint` / `typecheck` / `npm test` green — **515 tests**, coverage **99.6%** (gate 80%); new modules at 100% lines.

## Deferred (tracked)

- AW event fetch + `UsageSample` normalisation/rollups → **#88**.
- Admin-facing burndown surfacing → **#62**.
- Wiring the scheduled job into app bootstrap (needs the server SSH keypair from the #39 Phase-4 first-run + a background-jobs lifecycle the app doesn't have yet) → small follow-up once #39 lands; delivered here as composable, fully-tested units (mirrors how the timekpra wrapper #83 landed un-wired).

Closes #86


BenSeymourODB:
First-pass review (automated, deep-read of the full diff against `main`) — summary:

**License boundaries — verified PASS.** `withPortForward` is pure TCP forwarding via `ssh2.forwardOut` (no GPL linkage, no vendored source); ActivityWatch is reached only over HTTP REST on the loopback end of the tunnel. Loopback-only bind is enforced structurally — `LOOPBACK_HOST = "127.0.0.1"` is the only `listen()` bind address and `PortForwardOptions` exposes no bind-host knob, so the unauthenticated `aw-server` API can't be network-exposed. No new dependencies.

**One substantive finding, fixed in da992d4:** teardown destroyed the accepted sockets but not the `forwardOut` channels — `socket.pipe(channel)` doesn't propagate a socket `destroy()` to the SSH channel, so a channel could linger half-open on the pooled connection across passes, contradicting the documented "never leaks a listener, socket, or channel" guarantee. Now both sockets and channels are tracked and destroyed in the `finally`, with a regression test that holds a connection open across the window and asserts the channel is destroyed.

**Verified correct (no change needed):** concurrency bound in `mapWithConcurrency` (shared-cursor increments safe under single-threaded JS, no off-by-one; clamps `limit` to `[1, length]`); error taxonomy (`SshUnreachableError` → non-punitive `skippedOffline` gap, all else isolated as `failed` without aborting the pass); scheduler `protect` + awaited async handler + `catch` routing; type strictness (no `any`/`as`/`@ts-ignore`/`!`). One flagged nit (dropping the `{ ...result }` spread) was **not** applied — the spread is load-bearing: a TS interface has no implicit index signature, so `result` isn't directly assignable to the logger's `Record<string, unknown>` param.

Gate green after the fix: `format:check` / `lint` / `typecheck` / 516 unit tests, coverage 99.6% (gate 80%).

---

Issue #210, opened 2026-06-20
Tooling: CI auto-fix for Drizzle migration snapshot collisions (Slice 2 of #199)


BenSeymourODB:
## What

Slice 2 of #199, split out so the low-risk local helper (Slice 1) could ship on
its own. Slice 1 added `npm run db:rebase` (a local, staged-not-committed helper
that drops branch-only migrations + snapshots, trims the journal, regenerates
off the merged base, with safety guards). This issue is the **opt-in CI
auto-fix** that runs that helper automatically.

## Scope

A workflow on `claude/**` branches that, on PR `synchronize`, detects the
drizzle-kit snapshot **parent-collision** failure specifically, runs the Slice-1
`db:rebase` logic, and pushes the regenerated migration back to the PR branch.

Reuse the Slice-1 orchestrator (`server/scripts/rebase-migrations.ts`) — its
git/fs/script seams already make it scriptable; the workflow only supplies an
auto-commit + push step and a trigger.

## Concerns to design for (these are the real cost, not the code volume)

- **Trigger narrowly.** Only act when the `migrations` round-trip check is red
  *for the collision reason* — never on an unrelated failure, and never mid
  human review.
- **Honour the Slice-1 safety guard.** Never auto-regenerate a hand-edited /
  multi-migration branch (Slice 1 refuses without `--force`; the workflow must
  **not** pass `--force`). On a guarded refusal, comment for human attention
  instead of pushing.
- **Loop prevention.** The auto-commit must not retrigger itself — skip when
  `HEAD` is the bot's regen commit, or gate on a `[skip-regen]` marker.
- **Token scope.** The default `GITHUB_TOKEN`'s pushes don't re-trigger
  workflows; decide whether a write-scoped app/token is needed so the
  round-trip re-runs after the auto-fix (we likely *want* the re-run here).
- Leave a clear audit trail (commit message + PR comment) of what was
  regenerated and why.

## References

- Slice 1: #199 (`npm run db:rebase`, `server/scripts/rebase-migrations.ts`,
  `docs/testing.md`).
- Filename convention this builds on: #133.
- No license-boundary or tamper-resistance surface.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

[...]

Issue #199, opened 2026-06-19
Tooling: auto-rebase a branch's Drizzle migration after a merge from main


BenSeymourODB:
## Problem

When `main` advances while a feature branch carries its own Drizzle migration,
merging `main` in produces conflicts that the `migrations` CI job
("Drizzle migration round-trip") catches — but resolving the **journal `idx`**
is not enough. Each migration carries a **snapshot with a parent pointer**, and
two branches that both branched off the same snapshot generate migrations whose
snapshots claim the *same parent*. After merge that surfaces as:

```
[…_snapshot.json, …_snapshot.json] are pointing to a parent snapshot:
…/snapshot.json which is a collision.
```

The reliable manual fix is: **drop the branch-only migration(s) + their
snapshots, trim the journal entries, and re-run `npm run db:generate` against
the merged base** so the new migration chains off the latest snapshot. This bit
PR #184 (audit log, #85) twice — once on the merge of merged-main work, and
again after #181 (user-groups) landed ahead of it.

The timestamp-prefix convention (#133) keeps the migration *filenames* from
colliding, but not the snapshot *parent chain*. This issue tracks automating the
fix so contributors (and the implement-issue agent) don't have to know the
drizzle-kit internals.

### Slice 1 — local `db:rebase` helper (do first; low risk)

A `scripts/rebase-migrations.sh` / `npm run db:rebase` that:

1. Refuses to run if the working tree has unresolved conflict markers (the
   `schema.ts`/`enums.ts`/`_journal.json` source merge must be resolved first —
   regen fixes the migration artifacts, not the source conflict).
2. Computes branch-only migrations:
   `git diff --name-only origin/main...HEAD -- 'server/drizzle/*.sql'`.
3. **Safety guard:** refuse (or require `--force`) when
   - there is more than one branch-only migration (regen collapses them into a
     single cumulative diff — lossy), or
   - a branch-only migration is **not** byte-identical to what regen would
     produce, i.e. it was hand-edited or carries custom/data SQL. This repo
     already has a hand-fixed migration precedent (the #146 recurrence
     recreate, locked by `tests/policy/migrations.test.ts`), so blind
     regeneration is a real footgun.
4. `git rm` the branch-only `.sql` + their `meta/*_snapshot.json`; trim their
   `_journal.json` entries.
5. `npm run db:generate` then `npm run db:check`.
6. Leave the result **staged, not committed** — the human/agent reviews and
[...]

Issue #146, opened 2026-06-17
Schema: reserve recurrence + date-scoping columns per ADR 0004 (degenerate default = always-on)


BenSeymourODB:
## What

Adopt the schedule/budget schema shape implied by the recurrence + date-scoping ADR (#139) **now, in Phase 2**, so the foundational tables are right *before* the policy CRUD (#51), the editors (#53/#63), and the Phase 4 enforcement push are built against them. Reserving the columns is cheap; migrating them later — after CRUD, editors, and transport already depend on the old shape — is not.

## Why Phase 2

The landed schema (#48) models only uniform policy (flat `seconds_allowed`; free-text `cron_or_window`; `exceptions` with only `expires_at`). If recurrence/date-scoping arrives later, it forces a migration on the most central tables plus a rewrite of the resolver, editors, and push. This issue makes the model forward-compatible up front, without implementing the behaviour yet.

## Scope

- Per ADR 0004 (#139): add the chosen recurrence representation to `Schedule` (replacing the undefined free-text `cron_or_window`) and `effective_from` / `effective_to` to `Exception` (and any date-scoped rule).
- **Degenerate default = always-on:** a row with no recurrence and no effective window behaves exactly like today's uniform rule, so Phase 2 CRUD and (later) enforcement work unchanged while the columns sit ready.
- Drizzle migration + zod DTOs + `CHECK`/validation consistent with `enums.ts`.
- **No** resolver or UI behaviour change here — that lands in #140/#143 (Phase 4) and #141/#142 (Phase 13).

## Dependencies

- **Blocked by:** #139 (the ADR fixes the representation).
- **Blocks:** #51 (CRUD builds on these columns), #140, #141, #142, #143.

BenSeymourODB:
🤖 implement-issue claiming this for the current session.

Blocker #139 (ADR 0005) is **closed/accepted**, so this schema reservation is unblocked. Scope this run to exactly what ADR 0005 → "Reserved schema shape" fixes:

- `schedules`: replace the free-text `cron_or_window` with the structured recurrence (`recurrence_days` 7-bit ISO-weekday mask, `recurrence_start_minute`/`recurrence_end_minute` minutes-of-day) + `effective_from`/`effective_to`, all nullable, with `CHECK` constraints (both-or-neither minutes, `start < end`, days `1..127`, `effective_from < effective_to`).
[...]

PR #211, opened 2026-06-20
feat(scripts): db:rebase to re-chain a branch's Drizzle migration (#199)

BenSeymourODB:
## Summary

- Adds `npm run db:rebase` (**Slice 1** of #199): a dev-only CLI that recovers
  from the drizzle-kit snapshot **parent-collision** that happens when `main`
  advances while a branch carries its own migration. It drops the branch-only
  migration(s) + their snapshots, trims the `_journal.json` entries, re-runs
  `db:generate`/`db:check`, and leaves the result **staged, not committed**.
- **Safety guards** (from the issue): refuses on an unresolved merge (unmerged
  paths or leftover conflict markers); refuses — unless `--force` — when regen
  would be lossy (>1 branch-only migration, or a hand-edited / custom-SQL
  migration regen does not reproduce, restoring artifacts on that detection).
- Pure helpers + an orchestrator behind injected git/fs/script seams (mirrors
  the offline-queue pattern), unit-tested with in-memory fakes — no live git or
  drizzle-kit.

## Plan

Linked plan: `.claude/plans/issue-199-db-rebase-migrations.md`
Roadmap: `docs/roadmap.md` → Phase 11 (cross-cutting dev tooling).

## Design notes

- **No new dependency.** Runs via Node 22's built-in `--experimental-strip-types`
  (`node --experimental-strip-types scripts/rebase-migrations.ts`), so the logic
  is testable TypeScript without adding `tsx`.
- **Out of the Docker image.** The tool lives in `server/scripts/` (outside
  `src/`); `tsconfig.build.json` only includes `src/**`, so it never ships in the
  runtime image. It is added to `tsconfig.json` `include` + ESLint so the gate
  still covers it; coverage `include` stays `src/**`, so the CLI wiring doesn't
  skew the 80% gate.
- Honours the hand-edit precedent (`tests/policy/migrations.test.ts`, the #146
  recurrence recreate) — that's exactly the case the guard refuses to clobber.

## License-boundary note

N/A — no transport or packaging changes. Plain git + filesystem + npm-script
orchestration; no GPL imports, no GPL binary added to the image. The tool is
explicitly excluded from the Docker image.

## Deferred

Slice 2 (opt-in CI auto-fix on PR `synchronize`: narrow triggering, loop
prevention, write-scoped token, never `--force`) is tracked in **#210**. This PR
is **Part of #199**; #199 stays open for Slice 2.

## Test plan

- [x] `tests/scripts/rebase-migrations.test.ts` — 24 unit tests: pure helpers
      (conflict markers, journal trim+renumber, diff parsing, path derivation,
      SQL equivalence, argv) and the orchestrator (refuse on unmerged/markers,
      no-op, >1-migration guard + `--force`, happy single rebase, hand-edit
      detect+restore, `--force` overrides).
- [x] Full gate green from `server/`: `format:check`, `lint`, `typecheck`,
      `npm test` (816 tests pass, coverage 98.86% ≥ 80%).
- [x] CLI smoke test: `npm run db:rebase` on this branch reports the no-op path.

Part of #199

---
_Generated by [Claude Code](https://claude.ai/code/session_01AGoWw8SDfH1nd8mZvanoAV)_


BenSeymourODB:
Ran a first-pass self-review (subagent) before marking ready. Two findings worth a second commit (a46a1f9):

- **Abort/restore robustness:** the hand-edit/refusal path now restores the original `.sql` + snapshot + journal from an **in-memory capture** taken before any mutation, rather than `git checkout` (which only restores the index version and could silently drop a staged-but-uncommitted or freshly-generated hand edit). `db:generate` / `db:check` failures are now caught and restore the tree too, so an expected refusal never throws and never leaves a half-applied state.
- **Hardening:** `_journal.json` validated with zod (no unchecked `as`); SQL-equivalence made order-insensitive (sorted statement set split on `--> statement-breakpoint`); warn on a missing snapshot; error on unknown CLI flags.

Test count went 24 → 32 (failure-restore paths, journal/multi-file conflict scan, snapshot-missing warning, order-insensitive equivalence, unknown-flag). Full gate green locally (824 tests, 98.86% coverage). CI running now.

---

Issue #105, opened 2026-06-16
Admin UI: /admin/notifications (sound profile, master enable/disable, grace override)


BenSeymourODB:
## What

The `/admin/notifications` surface for editing the per-user `NotificationPolicy`: per-user sound profile, master enable/disable, and grace-period override. Per Phase 8b in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).

## Scope

- Per-user editor for `enabled`, `sound_profile`, `grace_seconds`, and cadence overrides.
- Writes through `/api/*` → `NotificationPolicy` (#103); changes propagate to clients via `policy.changed`.

## Notes

- Phase 8b "Admin UI under `/admin/notifications`" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- The notification + filter policy block also appears in the policy-editor mock [`design/admin/policy-editor.html`](../blob/main/design/admin/policy-editor.html) (PR #60) — keep them consistent.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

**Planned scope (a complete, end-to-end slice):** the `/admin/notifications` admin surface — a per-user editor for the existing `NotificationPolicy` (`enabled`, `soundProfile`, `graceSeconds`, and per-budget cadence overrides). The backend is already in place (`GET/PUT/DELETE /api/users/:userId/notification-policy` with `policy.changed` push propagation, #104), so this PR is the consuming UI:

- `frontend/src/lib/api/notifications.ts` — typed client over the existing endpoints.
- `frontend/src/lib/views/NotificationsView.svelte` — per-user editor with reset-to-defaults (DELETE) and master enable/disable.
- Nav wiring into the admin `AppShell`, contract-type re-exports, and view tests consistent with the existing view-test convention.

Kept consistent with the notification block in the `design/admin/policy-editor.html` mock (PR #60). No transport/license-boundary changes — type-only frontend consumption of the existing `/api`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #103, opened 2026-06-16
pct-client-agent: per-user notifications, sound, warning cadence, grace countdown, force-close


BenSeymourODB:
## What

The per-user `pct-client-agent` service (TypeScript, `systemd --user`) that renders the supervised-user-facing experience described in [`docs/client-notifications.md`](../blob/main/docs/client-notifications.md). Mocked in PR #60 — [`design/client/notifications.html`](../blob/main/design/client/notifications.html).

## Scope

- Notification + sound rendering via the desktop's own tools (`notify-send` / `gdbus`, `canberra-gtk-play`).
- Locally computed **warning cadence** (15 / 5 / 1 minute rules) against the cached budget, so warnings tick even between server pulls and while briefly offline.
- **Grace-period countdown** on `enforce.force_close`, then the per-app **force-close** (the kill the server delegates to the agent).
- Receives events from the bridge over `AF_UNIX`; holds the locally cached budget + `NotificationPolicy`.

## Notes

- Phase 8b "`pct-client-agent` per-user service" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Framing must be non-punitive ("time left", grace, not "blocked/violation") per the design principles in `design/README.md`.
- The cached read-only budget here is also what the Phase 12 "My Time" dashboard (#61) reads.

BenSeymourODB:
## Note: notification action-button behavior is out of scope here — tracked in #337

This issue's Scope is **render-only** (cadence, grace countdown, force-close), but the mockup it adopts as its visual spec — [`design/client/notifications.html`](../blob/main/design/client/notifications.html), from PR #60 — shows interactive toast buttons (the "Save & quit now" button, plus a styled-but-unused primary `.tbtn.go` action variant). The behavior behind those buttons was never specified here.

That gap is now tracked in **#337** (pulled into Phase 8b from the NDWC reward-redemption roadmap, [`docs/ndwc-reward-redemption-roadmap.md`](../blob/main/docs/ndwc-reward-redemption-roadmap.md)): action-button rendering + a generic, capability-gated client→server action frame. **Treat the buttons in the mockup as intentionally inert under #103 until #337 lands.**

Two clarifications on what "an action button" can actually do:

- **Local vs. remote.** Some actions are local to the agent and need no server round-trip (e.g. force-closing the user's *own* process); others must call home (redeem time, ask a parent for more). #337 covers both — the local handler and the generic client→server frame.
[...]

Issue #100, opened 2026-06-16
events module: GET /api/events/stream WebSocket with per-client bearer auth + reconnect


BenSeymourODB:
## What

The server `events/` module exposing `GET /api/events/stream` (WebSocket) that pushes events to enrolled clients, per [`docs/client-notifications.md`](../blob/main/docs/client-notifications.md) and the transport table in [`docs/architecture.md`](../blob/main/docs/architecture.md).

## Scope

- WebSocket endpoint with **per-client bearer-token** auth (the token issued at enrolment, #77).
- Client-initiated, long-lived, outbound connection model (works through home NAT); defined reconnect semantics.
- Event taxonomy + zod-typed payloads: `grant.applied`, `policy.changed`, `enforce.force_close`, `enforce.session_lock`, `lockout.cleared`.
- Per-client fan-out / addressing so an event reaches the right client's bridge.

## Notes

- Phase 8b first bullet in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- This is the channel the Phase 8 enforcement emit (#above), the Phase 8c lockout flow, and Phase 10 grants all publish onto. Build it first within 8b.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Selection trace: lower-phase eligible work is unavailable this run — #85/#124-foundation/#164/#154 are in-flight as open PRs (#184/#181/#180/#179); the phase-2/3/4 UI items (#53/#59/#63/#64/#81/#183) are blocked on the not-yet-built `/admin` shell (#53); #157's live `timekpra`-over-SSH round-trip can't be exercised in this unattended sandbox; #140 (Phase 4), #178 (Phase 5) and #94 (Phase 7) were each claimed <6h ago by concurrent sessions; the Phase-6 Ansible playbooks (#90–#93) have no `client/ansible/` scaffold or Molecule harness yet; and #98 (Phase 8) has a forward dependency on the Phase-10 grants overlay / effective-budget computation that isn't built. #100 is unclaimed, backend-only, has all its dependencies met (enrolment + per-client token #77 merged), is explicitly the "build it first within 8b" foundation, and contends with none of the in-flight PRs.

Scope for this session (server-side, fully unit-verifiable in-process):
- `events/` module exposing `GET /api/events/stream` (WebSocket) with **per-client bearer-token auth** (the enrolment token).
[...]

Issue #48, opened 2026-06-16
Drizzle policy schema + first real table migration


BenSeymourODB:
## What

Define the policy-store tables in `server/src/policy/schema.ts` on top of the
empty Phase-1 scaffold (#9), and generate the first real drizzle-kit migration
under `server/drizzle/`. This is the data foundation for everything else in
Phase 2 — today `schema.ts` is deliberately `export {}` and `drizzle/` holds
only an empty journal.

Entities per `docs/architecture.md` → "Policy model":
`User`, `Client`, `UserOnClient`, `Activity`, `ActivityGroup` (+ the
Activity↔Group many-to-many), `Budget`, `Schedule`, `Exception`, `UsageSample`,
`Grant`, `IntegrationToken`, `NotificationPolicy`.

## Notes / shape

- Mirror the sketch in `architecture.md`. Model the enums (`scope`,
  `window`, `Activity.kind`, `Schedule.action`) as `text` columns with a
  check constraint or a shared zod-derived union so the API layer can reuse
  the same source of truth.
- **`Grant` is an immutable ledger.** Encode it structurally now even though
  the grant *endpoints* are Phase 10: `source_ref` carries a **unique** index
  (the integrator idempotency key) and revocation is a `revoked_at` timestamp,
  never an in-place edit. Per-day effective budget = policy + Σ(active grants)
  is computed later, but the constraints belong here.
- `IntegrationToken` stores only `hashed_secret` (no plaintext).
- **Timezone column is gated on the #17 decision:** add `User.tz` only if #17
  lands on Option B. Land #17 first.
- Run `npm run db:generate` to emit the migration; commit it (journal
  included). `npm run db:migrate` applies cleanly against a fresh DB,
  `db:check` reports no drift, and re-apply is a no-op — extend the existing
  `server/tests/policy/migrations.test.ts`.

## Acceptance criteria

- [ ] All entities above defined with FKs and `user_id`-leading composite
      indexes on the hot query paths.
- [ ] Migration generated and committed under `server/drizzle/`; `db:check`
      green; re-apply-is-a-no-op test passing.
- [ ] `source_ref` uniqueness and `Grant` immutability (revoked_at, no edit)
      encoded at the schema level.
- [ ] Coverage gate (80%) stays green.

## References

- `docs/architecture.md` → "Policy model"
- #9 (drizzle scaffold this builds on), #17 (timezone → `User.tz`), #34
  (`DATABASE_URL` contract)
[...]

PR #213, opened 2026-06-20
feat(policy): NotificationPolicy — persist + push to the client (#104)

### LINKED ISSUE #104: NotificationPolicy: persist in the policy store + push to the client
## What

Add the `NotificationPolicy` entity to the policy store and push it to the client with the rest of policy, so the agent's behaviour (enable/disable, sound profile, grace seconds, cadence overrides) is server-controlled and cached client-side. Per the schema sketch in [`docs/architecture.md`](../blob/main/docs/architecture.md) and [`docs/client-notifications.md`](../blob/main/docs/client-notifications.md).

## Scope

- Drizzle schema + migration for `NotificationPolicy (user_id, enabled, sound_profile, grace_seconds, cadence_overrides_json)`.
- Include it in the policy push so the bridge/agent receive and cache it (delivered via `policy.changed` on the event stream).
- zod DTOs in `server/src/api/`.

## Notes

- Phase 8b "`NotificationPolicy` persisted in the policy store and pushed to the client" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Consumed by the agent (#102) and edited from the admin notifications page (#below).

BenSeymourODB:
## Summary

- Builds out the per-user **`NotificationPolicy`** (Phase 8b): a corrected schema + migration, repository access, zod DTOs, a `/api` CRUD surface keyed by user, and wiring into the existing Phase-2 policy-push stub so a change fans out to the user's linked clients.
- The `notification_policies` table was scaffolded in #48 with **placeholder values that contradicted the authoritative docs** (`sound_profile` defaulted to `"default"` with no enum CHECK; `grace_seconds` defaulted to `60` with no max). This aligns the entity with `docs/client-notifications.md` → "Configuration knobs" (`off|subtle|prominent`, default `subtle`; grace default `15`, range `0–60`) and `docs/architecture.md` → "Policy model".
- The two `schema.test.ts` assertions that locked in the placeholder defaults are updated to the corrected, doc-faithful defaults — correcting placeholder scaffolding to the spec, not weakening a behavioural test.

## Plan

Linked plan: `.claude/plans/notification-policy.md`
Roadmap: `docs/roadmap.md` → Phase 8b ("`NotificationPolicy` persisted in the policy store and pushed to the client").

## What's here

**Data layer** (`feat(policy)`)
- `enums.ts`: `soundProfile` enum (`off | subtle | prominent`) as a single-source tuple + zod schema.
- `policy/notification.ts`: shared grace bounds (0–60), documented defaults, `defaultNotificationPolicy()` — one source the storage CHECK and the DTOs both read (the `recurrence.ts` pattern).
- `schema.ts`: `sound_profile` enum-typed + CHECK, default `subtle`; `grace_seconds` default `15` + `between 0 and 60` CHECK.
- `repository.ts`: `get` / `upsert` (insert … on conflict update by the `user_id` PK) / `delete`.
- One timestamp-prefixed migration (table-recreate — SQLite can't alter a CHECK/default in place); `db:check` clean; table set unchanged so `migrations.test.ts` holds.

**API + push** (`feat(api)`)
- `GET /api/users/:userId/notification-policy` — persisted row, or the documented defaults when unset (a user always *has* an effective policy).
- `PUT …` — upsert; emits a user-scoped `notification.upserted` push.
- `DELETE …` — revert to defaults; emits `notification.deleted`; `404` when there's no custom policy (not a silent 204).
- `transport/stub.ts`: `notification.upserted` / `notification.deleted` reasons (user-scoped, like `budget.*`); the offline-queue adapter is unchanged (coalesce key stays `user:<id>`).

## Deferred (downstream consumers, already tracked)

- Wire delivery via `policy.changed` on the event stream — **#100** (PR #193).
- Admin `/admin/notifications` editor — **#105**.
- Agent consumption (cache + cadence) — **#103**.

The push payload + DTOs land here so those are thin follow-ups. No new issue needed — each deferred piece already has a roadmap ticket.

## License-boundary note

No GPL imports and no GPL binaries added to the image — plain TypeScript + zod + Drizzle (Apache-2.0) + better-sqlite3 (MIT). This PR establishes no new transport and collapses no process/network boundary; it only extends the in-process policy store and the existing logging push-stub seam.

## Test plan

- [x] `npm run format:check` / `lint` / `typecheck` clean
- [x] `npm test` — 811 unit tests pass; coverage gate (80%) held; new files at 100%
- [x] `npm run db:check` clean; migration round-trip + naming tests pass
- [x] Schema: corrected defaults; reject invalid `sound_profile`; reject `grace_seconds > 60`
- [x] Repository: upsert insert + partial-merge + cadence-clear + delete + cascade
- [x] Routes: GET-defaults / upsert / partial-merge / revert-to-defaults / 404s / validation / admin guard
- [x] Push stub: `notification.*` fan-out to linked clients; no push for reads/rejected writes

Closes #104

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vzi2siGrACd2iDbT5xAXdn)_


BenSeymourODB:
Merged latest `main` and resolved conflicts (group-targeting #182 / OS-account-fields #230 landed alongside this).

Two notes for review:

- **Migration regenerated** — the branch-only `…_clear_mimic` migration's parent snapshot collided once `main` advanced (the #199 scenario), so I dropped it and regenerated `…_remarkable_tombstone` against the merged schema; it now chains off `main`'s latest snapshot. `db:check` is clean and the migration round-trip/naming tests pass.
- **One cross-cutting touch outside #104's original scope:** `enforcement/evaluate.ts` (from #98) hardcoded its "no notification-policy row" grace fallback as `60` with a comment claiming it mirrored the schema default. This PR corrects that schema default to the documented **15** (`docs/client-notifications.md`), so I pointed the fallback at the new single-source `DEFAULT_GRACE_SECONDS` constant and updated the two enforcement test fixtures (default-path `60→15`; in-policy `120→45`, since 120 now violates the new documented `0–60` grace CHECK). Test intent is preserved, not weakened — happy to split this into a follow-up commit/issue if you'd prefer it isolated.

Full gate green on the merged tree: `format:check`, `lint`, `typecheck`, 1230 unit tests + coverage gate, `db:check`.

---

server/src/api/policy/routes.ts
```diff
@@ -1311,4 +1315,81 @@ export function registerPolicyRoutes(scope: FastifyInstance, push?: PolicyPushSt
...
+      );
+      pushStub.push(
+        userPushCommands(
+          "notification.upserted",
+          userId,
+          repo.listUserClientIds(scope.db, userId),
+          {
+            enabled: row.enabled,
+            soundProfile: row.soundProfile,
+            graceSeconds: row.graceSeconds,
```
BenSeymourODB, line 1393: Non-blocking: the `notification.upserted` push payload carries `enabled`/`soundProfile`/`graceSeconds` but omits `cadenceOverrides`, even though the cadence override is part of the policy "pushed to the client with the rest of policy" (and is persisted by the upsert just above). Since this is the log-only stub and the real wire delivery is `policy.changed` (#100), it's harmless today — but when #100/#103 consume this, the payload shape should include `cadenceOverrides` so the client cache isn't missing a field the user just changed. Worth either including it now or leaving a `// TODO(#100): include cadenceOverrides in the pushed payload` so it isn't forgotten.

---

server/drizzle/20260622191355_remarkable_tombstone.sql
```diff
@@ -0,0 +1,16 @@
...
+	`enabled` integer DEFAULT true NOT NULL,
+	`sound_profile` text DEFAULT 'subtle' NOT NULL,
+	`grace_seconds` integer DEFAULT 15 NOT NULL,
+	`cadence_overrides_json` text,
+	FOREIGN KEY (`user_id`) REFERENCES `users`(`id`) ON UPDATE no action ON DELETE cascade,
+	CONSTRAINT "notification_policies_sound_profile_check" CHECK("__new_notification_policies"."sound_profile" in ('off', 'subtle', 'prominent')),
+	CONSTRAINT "notification_policies_grace_check" CHECK("__new_notification_policies"."grace_seconds" between 0 and 60)
+);
+--> statement-breakpoint
+INSERT INTO `__new_notification_policies`("user_id", "enabled", "sound_profile", "grace_seconds", "cadence_overrides_json") SELECT "user_id", "enabled", "sound_profile", "grace_seconds", "cadence_overrides_json" FROM `notification_policies`;--> statement-breakpoint
```
BenSeymourODB, line 13: Migration-safety note (non-blocking, almost certainly safe today): this table-recreate copies existing rows via `INSERT … SELECT`, and the new table carries the tighter `sound_profile in ('off','subtle','prominent')` and `grace_seconds between 0 and 60` CHECKs. Any *pre-existing* row holding the legacy scaffolded default `sound_profile = 'default'` (or a `grace_seconds > 60`) would fail this copy with a CHECK violation and abort the migration.

In practice this is safe because no write path to `notification_policies` existed before this PR (#104 is what adds the CRUD), so any real DB has **zero** rows here and the SELECT copies nothing. Flagging only so it's a conscious call rather than an accident — if you want belt-and-suspenders, the copy could coerce a stale `'default'` → `'subtle'`, but I wouldn't bother given there can't be rows yet.

---

BenSeymourODB (commented):
## In-depth review (automated, scheduled session)

Reviewed the full diff against current `main`, checked out the branch, **reproduced the merge conflict, resolved it the intended way, and re-ran the entire quality gate on the merged tree.** This is a clean, doc-faithful slice — the data layer, API, and push wiring are all well-factored and the single-source-of-truth pattern (`policy/notification.ts` feeding both the storage `CHECK` and the DTO validation) is exactly right. Leaving this as **Comment, not Approve**, for one reason only: it currently doesn't merge (one conflict, detailed below). Everything else is green.

### ✅ License boundary — compliant
N/A as the PR claims: plain TypeScript + zod + Drizzle (Apache-2.0) / better-sqlite3 (MIT). No GPL import, no transport/REST boundary introduced or collapsed, no GPL binary added to the image. It only extends the in-process policy store and the existing logging push-stub seam (`transport/stub.ts`). `license-guard` unaffected.

### ✅ The modified tests are corrections, not weakenings — confirmed
I checked this specifically because the skill flags test edits. Both changes hold up:
- **`schema.test.ts`**: the old assertions (`soundProfile === "default"`, `graceSeconds === 60`) locked in *placeholder* values from the #48 scaffold that contradicted the authoritative doc. `docs/client-notifications.md` → "Configuration knobs" (line 256) and "Sound design" (line 178) state `sound_profile` default `subtle` (`off`/`subtle`/`prominent`) and `grace_seconds` default `15`, "0 disables the grace period; max 60". The new assertions match the doc, and the change *adds* two CHECK-constraint tests (off-enum profile, grace > 60) — net stronger coverage.
- **`evaluate.test.ts`**: the `120 → 45` edit is *required*, not cosmetic — the new `between 0 and 60` CHECK makes `120` an invalid stored value, so the test now uses an in-range value distinct from the default to still prove the grace is read from the policy, not the fallback. Good.

### ⛔ The one merge blocker — `server/src/enforcement/evaluate.ts`
The PR is `mergeable_state: "dirty"`. The only content conflict is in `evaluate.ts`: after this PR was opened, `main` independently grew a **local `const DEFAULT_GRACE_SECONDS = 60`** as the no-policy-row enforcement fallback. This PR removes that local const and imports the shared `DEFAULT_GRACE_SECONDS` (= `15`) from `policy/notification.ts`.

The correct resolution is to **keep this PR's side** (drop `main`'s local `= 60` const, keep the import), which makes the no-row grace fallback the documented `15`. I applied exactly that resolution locally and ran the full gate on the merged tree:

```
format:check ✓   lint ✓   typecheck ✓   db:check ✓
Test Files  112 passed (112)
Tests       1372 passed (1372)
Coverage    98.85%  (gate 80%)
```

So once rebased with that resolution, it's green. ⚠️ **Behavioural note worth a line in the PR body:** this means a user with *no* `notification_policies` row now gets a **15s** enforcement grace instead of `60s`. That's intended (doc-aligned) and it also reconciles a drift where `main` had landed `60` — but it is a real change to enforcement timing for the default case, so it's worth calling out explicitly.

### Nits (inline, non-blocking)
- The `notification.upserted` push payload omits `cadenceOverrides` (inline on `routes.ts:1364`).
- The migration's `INSERT … SELECT` table-recreate would reject a stale `sound_profile = 'default'` row under the new CHECK — safe today since no write path existed pre-#104, but flagged (inline on the migration).

Happy to take another pass once it's rebased.

---
_Automated in-depth review by Claude Code. I did **not** push changes — this session is pinned to a different branch, so the verified conflict resolution above is left for the author to apply. The full gate result quoted is from that resolution applied locally on top of current `main`._

---

server/src/api/policy/routes.ts
BenSeymourODB: Done — included `cadenceOverrides` in the `notification.upserted` push payload (`row.cadenceOverridesJson ?? null`) so the full effective policy travels "with the rest of policy", and updated the push-stub test to set a cadence override and assert it's carried. Pushed in 48b8354.

---

server/drizzle/20260622191355_remarkable_tombstone.sql
BenSeymourODB: Agreed it's safe, and leaving it as-is deliberately. Two reasons: (1) as you note, no write path to `notification_policies` existed before this PR, so any real DB has zero rows and the `INSERT … SELECT` copies nothing — the fresh-DB migration CI also proves the round-trip. (2) This `.sql` is drizzle-kit-generated; hand-editing it to add a `CASE … ELSE 'subtle'` coercion would make it byte-non-identical to what `db:generate` produces, which is exactly the "hand-edited migration" footgun the #199/#210 auto-rebase tooling treats as un-regenerable. So the belt-and-suspenders coercion would cost more than it's worth for a table that can't hold rows yet. Flagging-as-conscious-call: confirmed conscious. 👍

---

BenSeymourODB:
Thanks for the thorough pass. Status after this round:

- **Merge blocker resolved.** `main` had advanced again, so I re-merged it and resolved the `enforcement/evaluate.ts` conflict exactly the way you recommended — keep this PR's side, so the no-`notification_policies`-row enforcement grace fallback uses the single-source `DEFAULT_GRACE_SECONDS` (the documented **15s**) rather than `main`'s re-introduced local `60`. (Behavioural note already in the PR/commit: a user with no notification-policy row now gets a 15s enforcement grace, down from 60 — intended/doc-aligned.)
- **Nit 1 (cadenceOverrides):** fixed — now in the `notification.upserted` push payload + asserted in the push-stub test.
- **Nit 2 (migration stale-`'default'` coercion):** deliberate non-action — no rows can exist pre-#104, and hand-editing the drizzle-generated `.sql` would break regen-identity for the #199/#210 auto-rebase tooling. Explained inline.

Full gate green on the merged tree: `format:check`, `lint`, `typecheck`, `db:check`, **1372 unit tests** + coverage gate. Pushed as 48b8354; CI re-running.

---

Issue #90, opened 2026-06-16
Playbook: e2guardian per-UID filter groups + iptables OUTPUT redirect


BenSeymourODB:
## What

The Ansible playbook(s) that make web filtering actually take effect on a client: write e2guardian per-Linux-UID filter-group config and reload the daemon, and install the iptables OUTPUT-chain redirect that sends each supervised UID's traffic to e2guardian's proxy port. These are paired — the filter is inert without the redirect (see [`docs/architecture.md`](../blob/main/docs/architecture.md) "Enforcement responsibilities").

## Scope

- e2guardian: render per-UID filter groups from policy (`Activity` domain matchers / blocklists), write config, signal reload. **Config-file + signal only**, no code integration (license boundary).
- iptables: per-UID OUTPUT redirect to the e2guardian proxy port; idempotent rule management.
- Driven by the Ansible runner (#above); recorded in the audit log (#85).

## Notes

- Phase 6 "Playbooks for e2guardian filter groups, iptables OUTPUT redirect" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Per-website *time-window* swaps (config swap on schedule) build on this; can be a follow-up.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Plan for this PR (a meaningful, mergeable slice):

- **Server-side plan generation** (`transport/ansible/e2guardian`): derive a typed, zod-validated per-UID e2guardian filter plan from policy — the always-on domain/`domain_group` denies for each supervised user on a client (`deny` schedules targeting domain-kind activities, with no recurrence/date bounds), expanded to concrete domain matchers, with deterministic filter-group + listen-port assignment. Fully unit-tested (`playbook-generation.test.ts` per `docs/testing.md`).
- **Runner wiring** (`pushE2guardianFiltering`): pass the plan to the existing Ansible runner as `--extra-vars` and record the run in the audit log. Unit-tested against an injected runner.
- **Ansible playbook + templates** (`client/ansible/`, establishing the layout): render per-UID filter groups (banned site/url lists, group config) and the paired iptables OUTPUT redirect (`-m owner --uid-owner`) to e2guardian's per-user proxy port, idempotently, then reload. Validated with `ansible-playbook --syntax-check` + `ansible-lint`.

Deferred to follow-up issues (will file + link from the PR):
- **Per-website *time-window* config swaps** on a schedule (the issue notes this can be a follow-up).
[...]

Issue #216, opened 2026-06-20
e2guardian per-website time-window config swaps (recurring/date-scoped denies)


BenSeymourODB:
## What

Extend the e2guardian web filter (#90) to honour **time-varying** domain denies,
not just always-on ones. Today `buildE2guardianPlan` (`transport/ansible/e2guardian.ts`)
deliberately includes only the *always-on* `deny` schedules (every recurrence /
`effective_*` field null, per ADR 0005); a deny with a recurring weekday window
(e.g. "no YouTube 16:00–18:00 on weekdays") or a date range is skipped.

This issue adds the "config swap on schedule" half called out in
`docs/architecture.md` → "Enforcement responsibilities"
(*Per-website time-window | e2guardian config swap on schedule | Ansible +
systemd timer*) and noted as a follow-up in #90.

## To nail down

- **Mechanism**: a systemd timer on the client that swaps the active per-UID
  banned-site list between window/non-window variants, vs. the dashboard
  re-pushing the playbook at each window boundary (croner), vs. e2guardian's own
  time-based list support. Pick one and document it.
- Compose with the Phase-4 effective-policy resolver (#143) so "what is blocked
  for user U right now" has a single source of truth shared with enforcement and
  the burndown/coming-up views.
- Extend `e2guardianPlanSchema` + `buildE2guardianPlan` to carry windowed denies,
  and the playbook/templates to render them. Tests throughout.

## Depends on / relates

- Builds on #90 (always-on filter + playbook + plan contract).
- Resolver #143; recurrence model ADR 0005.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

**Mechanism decision (the issue's "pick one and document it"):** option 3 — **e2guardian's native time-based list support**, not a client systemd timer or a dashboard croner re-push. Rendering the recurring window directly into the banned-site list (`#time: <sh> <sm> <eh> <em> <days>` header on a `.Include`d list file) keeps enforcement purely *config-file + reload* — no new client-side scheduler, no clock-drift/re-push race, and no widening of the license boundary or the tamper surface. e2guardian evaluates the window itself at request time.

**Scoped this run to the recurring (weekday/time-window) case:**
[...]

Issue #215, opened 2026-06-20
e2guardian filtering: live Molecule test + port→group binding that makes it take effect


BenSeymourODB:
## What

The live integration test **and** the e2guardian directive set that makes per-UID
filtering actually take effect. The #90 PR landed the plan generation + the
playbook that renders per-group config/banned-site lists + the iptables OUTPUT
`--uid-owner` redirect, but **filtering is inert until the port→group binding is
wired** (see scope below). Both need a real e2guardian daemon to confirm, so they
land together here.

## Scope

1. **Make filtering take effect** (the gap #90 deferred):
   - Set `filtergroups = N` in `e2guardian.conf` to the managed group count so
     `e2guardianf2.conf`+ are actually loaded.
   - Wire the **per-listen-port → filter-group binding** (e.g. the `port`/`ports`
     auth plugin mapping each user's `listenPort` to their `filterGroup`), so
     traffic the iptables redirect sends to a user's port is filtered by *their*
     group rather than the permissive group 1.
   - Reconcile these into `pct-managed-listen.conf.j2` / `e2guardian-filtering.yml`.
2. **Molecule integration test** (`*.int.test.ts` / Molecule, `npm run test:integration`):
   - Container with a real e2guardian + iptables + a couple of supervised UIDs.
   - Apply the playbook with a representative `e2guardian` plan and assert: a
     banned domain is blocked for the targeted UID, an allowed domain is
     reachable, the iptables rule is present with the expected `--to-ports`, and
     a second apply is a no-op (idempotent).

## Why it was deferred from #90

The exact e2guardian per-port→group directive set can only be confirmed against a
real daemon, and a meaningful assertion needs a privileged container with
e2guardian + iptables (`nat`/`OUTPUT`, `-m owner`) — heavier infra than the unit
suite, and unavailable in the scheduled-run sandbox (no Docker daemon;
`ansible-playbook` can't run there). Same split as the live `timekpra`
round-trip (#207) vs. the unit-tested `timekpr` command generation, and the
AppArmor (#92) live-convergence deferral.

## References

- `client/ansible/playbooks/e2guardian-filtering.yml`,
  `server/src/transport/ansible/e2guardian.ts`, PR #217.
- `docs/architecture.md` → "Enforcement responsibilities"; `docs/testing.md`.

BenSeymourODB:
# Issue #215 — Molecule e2guardian-filtering readiness survey

[...]

Issue #178, opened 2026-06-19
Usage normalisation: richer activity matchers + domain/group-kind resolution


BenSeymourODB:
## What

Extend the AW → `UsageSample` normaliser shipped in #88 beyond its deliberate
v1 matcher semantics.

#88 resolves window events against `Activity` matchers by **case-insensitive
exact match** of the AW `app` against the `matcher`, and only for **`app`-kind**
activities. That was a conscious, documented v1 (the architecture doc leaves
"matcher" unspecified). This issue covers the richer cases it deferred:

- **Richer matching grammar** — substring / glob / regex matchers (e.g. matcher
  `chrome` matching the AW app `google-chrome`), with a clear precedence when
  several activities match one event. Probably wants a short ADR since it
  becomes the contract the admin's Activity editor (#53) writes against.
- **`domain` / `domain_group` kinds** — these match web requests, not window
  events, so they are sourced from web-proxy telemetry (e2guardian, Phase 6/7),
  not the AW window watcher. Define how domain usage becomes `UsageSample` rows
  once that telemetry exists.
- **`app_group`-kind resolution** — the client-expanded bundle kind, distinct
  from the `activities_to_groups` M2M that group *rollups* already use.

## Why deferred

#88 needed *a* deterministic resolution to produce samples and unblock the
burndown/enforcement readers; exact app-match is correct and predictable as a
floor. The richer grammar is a separable design decision that shouldn't gate the
data-layer landing.

## References

- Builds on #88 (`transport/activitywatch/normalise.ts`, `policy/usage.ts`).
- `domain*` work coordinates with the e2guardian playbooks (Phase 6, #90) and
  AdGuard (Phase 7).
- Activity matcher editor: #53 / #63.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Selection trace: lower-phase eligible work is unavailable this run — #140 was claimed <6h ago by a concurrent session; #85/#124-foundation/#164/#154 are in-flight as open PRs (#184/#181/#180/#179); #53/#59/#63/#64/#81/#183 are blocked on the not-yet-built `/admin` shell (#53); #157's live timekpr-over-SSH container can't be exercised in this environment (no Docker daemon). #178 is unclaimed, unblocked (#88 merged), backend-only, and contends with none of the in-flight PRs.

Scope for this session (the parts implementable + unit-verifiable now):
[...]

PR #217, opened 2026-06-20
feat(transport/ansible): e2guardian per-UID filter groups + iptables redirect (#90)

BenSeymourODB:
## Summary

Phase 6's e2guardian web filter — the server-side generation that drives it plus
the client playbook — landing alongside the sibling AppArmor (#92) and
ActivityWatch (#91) playbooks in `client/ansible/`.

- **Server-side plan generation** (`server/src/transport/ansible/e2guardian.ts`):
  a zod-validated per-UID filter plan (`e2guardianPlanSchema`) derived from
  policy — for each supervised user on a client, the **always-on `deny`
  schedules** targeting a `domain` Activity (directly or via a group), expanded
  to concrete domains, with deterministic filter-group + listen-port assignment.
  `pushE2guardianFiltering` hands the plan to the existing Ansible runner as
  `--extra-vars` and records the run in the audit log.
- **Client playbook** `client/ansible/playbooks/e2guardian-filtering.yml` + Jinja
  templates: render the per-UID filter group config + banned-site lists under the
  Phase-3-reserved `/etc/e2guardian/pct.d/`, and the **paired** iptables OUTPUT
  `--uid-owner` redirect to each user's listen port, idempotently, then reload.
- Reuses `listClientLinks` and the `ExtraVarValue` runner type that landed on
  `main`; the plan uses the `osUsername`/`osUserRef` link model (#230), matching
  the AppArmor plan.

## Plan

Linked plan: `.claude/plans/phase6-e2guardian-filtering.md`
Roadmap: `docs/roadmap.md` → Phase 6.

## ⚠️ Not yet live end-to-end — `Addresses #90`, does not close it

The playbook renders the per-group config + banned-site lists and installs the
per-UID iptables redirect, **but** the e2guardian directive set that actually
*selects* a filter group by listen port (`filtergroups = N` + the port→group
auth-plugin binding) is **not yet in place** — so redirected traffic is still
filtered by the permissive group 1 until that lands. That binding can only be
confirmed against a real e2guardian daemon, so it is folded into the Molecule
follow-up **#215** (scope expanded accordingly). This PR therefore **Addresses
#90** and keeps it open until #215 proves live filtering end-to-end, so the
roadmap doesn't show Phase-6 web-filtering as done while it can't take effect.

## License-boundary note

e2guardian is configured **only** by writing config files and signalling a
reload, driven by `ansible-playbook` as a subprocess — no in-process/GPL
linkage, no GPL binary added to the dashboard image (`CLAUDE.md` → "License
boundaries", rules 3 & 6). No Docker-image or packaging changes.

## Validation

- Merged current `main` in; `npm run format:check`/`lint`/`typecheck` clean,
  **1248 unit tests pass**, `e2guardian.ts` 100% lines, overall coverage 98.9%.
- `ansible-lint` (production profile, 0 findings) + YAML/Jinja parse on the new
  playbook + templates.

## Review follow-ups addressed

- Merged onto current `main` (was non-mergeable); reconciled to main's
  `ExtraVarValue` runner type and the existing `client/ansible/` scaffold +
  `listClientLinks`; plan adopts the `osUsername`/`osUserRef` model (#230).
- `filtergroups`/port→group binding gap called out in the README + folded into
  #215; `Closes` → `Addresses #90`.
- Nits: `listenPort` headroom comment, `target.port = 22` comment,
  `netfilter-persistent save` handler now `changed_when: false`.

## Deferred work (issues filed + linked)

- **#215** — live Molecule test **+** the `filtergroups`/port→group auth-plugin
  binding that makes filtering take effect (needs privileged Docker infra).
- **#216** — per-website time-window config swaps (recurring/date-scoped denies).
- `domain_group` (named-bundle) expansion stays with #178/#195.

Addresses #90


client/ansible/playbooks/templates/pct-managed-listen.conf.j2
```diff
@@ -0,0 +1,17 @@
...
+
+  NOTE: the exact e2guardian directive set for per-port filter-group selection
+  is pending the Molecule integration test (see the issue linked from the #90
+  PR). The dashboard-computed mapping below is the source of truth either way.
+-#}
+# Managed by pct e2guardian-filtering.yml (#90). Generated — do not hand-edit.
+# Base permissive proxy (filter group 1) listens on {{ e2guardian.proxyPort }}.
+{% for user in e2guardian.users %}
+# group {{ user.filterGroup }} ({{ user.linuxUsername }}, uid={{ user.linuxUid }}) <- listen port {{ user.listenPort }}
+filterports = {{ user.listenPort }}
```
BenSeymourODB, line 16: **Substantive gap (the one the template NOTE alludes to):** emitting `filterports = <port>` per user makes e2guardian *listen* on those ports, but on its own it does **not** bind a listen port to a filter group, and nothing here sets `filtergroups = N` in `e2guardian.conf`. As written:

1. e2guardian only loads `filtergroups` groups; `e2guardianf2.conf`+ are ignored unless `filtergroups` is bumped to the managed count.
2. Filter-group selection per connection is driven by an auth plugin (e.g. the `port`/`ports` auth plugin mapping listen-port → group), not by `filterports`. Without it, traffic redirected to a user's `listenPort` is still filtered by group 1 (the permissive baseline), so the per-UID denies never take effect.

So the iptables redirect + per-group config render, but the *filtering itself is inert* until the port→group binding lands. That's consistent with the deferral to #215, but it's a bigger gap than "directive details best-effort" implies — it's the mechanism that makes the feature work. Two asks:
- Make #215's scope explicitly include setting `filtergroups` and the per-port→group auth-plugin wiring (not just "apply the playbook").
- Consider whether `Closes #90` should be `Addresses #90`, keeping #90 open until #215 proves live filtering end-to-end (see the summary).

---

server/src/transport/ansible/e2guardian.ts
```diff
@@ -0,0 +1,279 @@
...
+      listenPort: proxyPort + 1 + index,
+      bannedSites,
+    });
+  }
+
+  return e2guardianPlanSchema.parse({ proxyPort, redirectPorts, users });
+}
+
+/** Shape the plan into the nested `--extra-vars` object the playbook reads. */
+function planToExtraVars(plan: E2guardianPlan): JsonObject {
```
BenSeymourODB: **Will conflict on rebase.** `main` has since widened the runner's `extraVars` from `JsonObject` to `Record<string, ExtraVarValue>` (the trial-merge of this branch onto current `main` hits a content conflict in `transport/ansible/index.ts` here). On rebase, drop this PR's `JsonObject` widening and have `planToExtraVars` return the type `main` now exports (`ExtraVarValue`/`Record<string, ExtraVarValue>`), updating the import from `./index.js` accordingly. No behaviour change — both are JSON-shaped — but the type needs to follow the one that landed.

---

server/src/transport/ansible/e2guardian.ts
```diff
@@ -0,0 +1,279 @@
...
+  const users: E2guardianUserFilter[] = [];
+  for (const link of listClientLinks(db, clientId)) {
+    const bannedSites = resolveBannedSites(db, link.userId);
+    if (bannedSites.length === 0) continue;
+    const index = users.length;
+    users.push({
+      linuxUsername: link.linuxUsername,
+      linuxUid: link.linuxUid,
+      filterGroup: FIRST_MANAGED_FILTER_GROUP + index,
+      listenPort: proxyPort + 1 + index,
```
BenSeymourODB, line 177: Minor: `listenPort = proxyPort + 1 + index` means a deployment that overrides `proxyPort` close to the top of the range could push a `listenPort` past 65535, and `e2guardianUserFilterSchema` would then reject it at `parse()` with a generic Zod error. Not reachable with the 8080 default, but a clearer guard (or a comment noting the assumed headroom) would make the failure mode obvious if someone sets a high custom `proxyPort`. Non-blocking.

---

BenSeymourODB (commented):
## In-depth review (automated, scheduled session)

Reviewed against current `main` (`71c3221`). This is a clean, well-documented slice and the **policy→plan derivation is excellent** — `buildE2guardianPlan`/`resolveBannedSites` are hermetic and thoroughly unit-tested (always-on-only filtering, group expansion, `domain_group` skip, deterministic UID-ordered group/port assignment, audit on every outcome). I'm leaving this as **Comment, not Approve**, because there are two things to settle before merge.

### ✅ License boundary — compliant
e2guardian is configured **only** by rendering config files under `/etc/e2guardian/pct.d/` + a reload handler, and iptables via the Ansible module — all driven by `ansible-playbook` as a subprocess. No e2guardian/Ansible code is linked, imported, or vendored; no GPL binary is added to the image. Matches `CLAUDE.md` → "License boundaries" rules 3 & 6. `INVENTORY_GROUP = "supervised"` correctly matches the playbook's `hosts: supervised`.

### ⛔ 1. Needs a rebase — it no longer merges cleanly
Trial-merging this branch onto current `main` produces three conflicts, because Phase-6 scaffolding landed after this PR was opened:
- `client/ansible/README.md` — **add/add**. `main` already established `client/ansible/` (it now has `molecule/`, `playbooks/activitywatch.yml`, `playbooks/apparmor-profiles.yml`, `ansible.cfg`, `README.md`). This PR's "establishing the layout" framing is stale — fold the e2guardian section into the existing README rather than replacing it.
- `client/ansible/ansible.cfg` — **add/add**. Prefer `main`'s; drop this PR's copy.
- `server/src/transport/ansible/index.ts` — **content**. `main` widened `extraVars` to `Record<string, ExtraVarValue>`; reconcile to that and drop this PR's `JsonObject` (see the inline note on `e2guardian.ts:175`).

After reconciling, re-run the full gate (`format:check`/`lint`/`typecheck`/`test`) — CI here was green on 2026-06-20 but against the pre-conflict base, so it needs to be re-validated on the merged tree.

### ⚠️ 2. The filtering is inert until the port→group binding lands (see inline on `pct-managed-listen.conf.j2:16`)
`filterports = <port>` makes e2guardian listen on the per-user ports, but nothing binds a listen port to its filter group and `filtergroups = N` is never set — so the rendered `e2guardianf2.conf`+ groups aren't loaded/selected and redirected traffic still hits the permissive baseline. The template NOTE flags this as deferred to #215, which is reasonable for a scaffolding slice, but it's the mechanism that makes the feature actually filter. Recommend:
- Expand **#215**'s scope to include `filtergroups` + the per-port→group auth-plugin wiring (not just "apply the playbook").
- Change `Closes #90` → **`Addresses #90`** and keep #90 open until #215 proves live filtering end-to-end, so the roadmap doesn't show Phase-6 web-filtering as done while it can't yet take effect.

### Nits (non-blocking)
- `pushE2guardianFiltering` records `target.port = 22`; fine today (`AnsibleHost` has no port), but worth a comment so it's revisited if custom SSH ports ever land.
- `listenPort` overflow with a high custom `proxyPort` (inline on `e2guardian.ts:166`).
- `Persist iptables rules` handler uses `changed_when: true`; `netfilter-persistent save` is a persistence side-effect, so `changed_when: false` would report more honestly. Trivial.

Happy to take another pass once it's rebased.

---
_Automated in-depth review by Claude Code. I did not push changes — this session is pinned to a different branch, so the remediation above is left for the author/maintainer._

---

BenSeymourODB:
Thanks — all four points addressed in the merge I just pushed (`89d5830`):

**1. Mergeability / rebase.** Merged current `main` in (rather than rebasing, to avoid a force-push on the published branch). Reconciled the three collisions you flagged, plus two more that landed since your review against `71c3221`:
- `index.ts` — dropped this PR's `JsonObject`; `extraVars`/`planToExtraVars` now use main's `ExtraVarValue`.
- `README.md` / `ansible.cfg` — folded the e2guardian section into main's existing `client/ansible/README.md`; took main's `ansible.cfg`; dropped my `inventory.example.ini`.
- Also: `main` added its **own** `listClientLinks` (ordered by user id) and renamed the link columns to `osUsername`/`osUserRef` for Windows support (#230). Dropped my duplicate helper, and the plan DTO/playbook/templates now use `osUsername`/`osUserRef` (matching the AppArmor plan). The deterministic group/port assignment now follows `listClientLinks`' user-id order.
- Re-ran the full gate on the merged tree: 1248 tests pass, typecheck/lint/format clean, `ansible-lint` production profile green.

**2. Filtering inert until port→group binding.** Agreed — this is the mechanism, not a directive detail. Done both asks: **#215** scope now explicitly covers `filtergroups = N` + the per-port→group auth-plugin binding (renamed/expanded), and the PR is now **`Addresses #90`** (keeps #90 open until #215 proves live filtering). The README e2guardian section calls the gap out plainly.

**3. `target.port = 22`.** Added a comment noting `AnsibleHost` carries no port today and to revisit if custom SSH ports land.

**4. `listenPort` overflow / `changed_when`.** Added a headroom comment on the port assignment in `buildE2guardianPlan`; switched the `netfilter-persistent save` handler to `changed_when: false`.

---

PR #312, opened 2026-06-26
Merge "Client Health" into a single "Clients" view (#305)

### LINKED ISSUE #305: Admin UI: merge "Client Health" into a single "Clients" view with the manual create behind an "Advanced enrolment options" pane
## Problem

The admin dashboard has two separate nav items — **Clients** and **Client Health** — and *both* invite the admin to "enrol a new client". This is confusing because the two entry points are not the same operation, and they aren't even doing the same thing:

- **Clients** view (`server/frontend/src/lib/views/ClientsView.svelte`) — an inventory CRUD table (hostname / SSH user / enrolled / last-seen / edit / delete). Its "Add client" form calls `POST /api/clients`.
- **Client Health** view (`server/frontend/src/lib/views/ClientHealthView.svelte`) — an operational status board (reachability, per-component health, offline queue). It *also* hosts the full **enrolment-token** flow: mint a one-time token → show the `install-client.sh` one-liner. It calls `POST /api/clients/enrolment-tokens`.

So the canonical onboarding flow lives under a tab called "Client Health" (the wrong place for it), while a *second*, lower-level "Add client" form sits under "Clients" and reads as a parallel way to "enrol". One concept, two homes, two different behaviours.

## What the enrolment workflow actually does (research summary)

I traced the full path end-to-end. There are **three distinct write paths** to the `clients` table, and they are genuinely different — the manual create is **not** redundant in capability, only in *labelling/placement*:

| Path | Endpoint | Auth | Creates | Bearer token | User↔client links | Versions |
|---|---|---|---|---|---|---|
| Mint token | `POST /api/clients/enrolment-tokens` | admin session | **token row only** (`enrolment_tokens`) | — | — | — |
| Claim / self-register | `POST /api/clients/enrol` | one-time bearer token (install script) | client row **+ links**, atomically | **set** (can use event stream) | created, bijection-enforced vs minted set | recorded at enrol |
| Manual create | `POST /api/clients` | admin session | client row only (`hostname` + `sshUser`) | **NULL** (cannot use event stream) | none | none |

Key facts that shape the design:

- **The client DB record is created at token-*claim* time (`/enrol`), not at mint time.** Minting only writes an `enrolment_tokens` row; the install script's `/enrol` call creates the `clients` row + `users_on_clients` links + sets the per-client bearer-token hash, all in one transaction (`server/src/policy/enrolment.ts` `consumeTokenAndEnrol`).
- **Enrolment does not require a pre-existing client record** — so the manual `POST /api/clients` is *not* a prerequisite for the token flow. They're independent.
- **A manually-created client is a degraded record**: no bearer token (so it can't authenticate to the Phase-8b event stream), no supervised-user links, no version inventory. Schema comment in `server/src/policy/schema.ts` confirms this is intentional, for "legacy clients / test machines / manual inventory".
- `GET /api/clients/health` (`server/src/api/clients/health-service.ts`) just *reads* existing client rows and joins them with the offline transport queue + an optional live SSH probe. It never creates records.

**Conclusion:** don't delete either capability. The token flow is the canonical front door and belongs front-and-centre; the manual create is a real-but-advanced escape hatch that should be demoted and clearly framed, not presented as a co-equal "enrol" path.

## Proposal: one "Clients" view

Merge `ClientHealthView` into `ClientsView` and drop the separate **Client Health** nav item. The unified view becomes the single place to **see, enrol, and manage** clients.

### 1. Unified client list (inventory + health in one table/cards)
Combine the inventory columns with the health columns, per client:
- hostname, SSH user, enrolled, last-seen *(from ClientsView)*
- reachability (online/offline/unknown), per-component health, queued/failed change counts *(from ClientHealthView)*
- per-row Edit / Delete actions *(from ClientsView)*
- expandable queued-actions detail *(from ClientHealthView)*

For a manually-created client (no bearer token), surface a small badge/affordance — e.g. "manual record — not fully enrolled" — so its degraded state is visible rather than silently identical to a real enrolment.

### 2. Primary action: "Enrol a client" → the token flow
Promote the enrolment-token flow (currently buried in Client Health) to the primary call-to-action on the Clients view: select supervised user(s) + OS username(s), optional hostname → **Generate enrolment token** → show the one-time token + copy-able `install-client.sh` one-liner. No behaviour change to the endpoints; this is a relocation.

### 3. "Advanced enrolment options" pane (collapsed by default)
Keep the existing `POST /api/clients` "Add client" (hostname + SSH user) form, but move it into a **collapsed-by-default** `<details>`-style "Advanced enrolment options" pane within the Clients view. Inside it, briefly state the trade-off: this creates a record manually without the enrolment exchange, so the client will have **no event-stream credential and no supervised-user links** until it goes through `/enrol`. This keeps the escape hatch for test machines / legacy clients without it competing with the normal path.

## Implementation sketch

- `server/frontend/src/routes/admin/+page.svelte` — remove the `client-health` entry from `navItems` (lines ~43–56) and the `{:else if activeView === "client-health"}` branch (lines ~124–147); the `clients` view now renders the merged component.
- `server/frontend/src/lib/views/ClientsView.svelte` — absorb the health data + enrolment-token UI:
  - fetch both `listClients()` and `listClientHealth()` (or have the list render primarily off `client-health`, which already carries hostname/enrolledAt/lastSeen, and fold in `sshUser`/edit-delete from the clients list);
  - add the "Enrol a client" token-mint section (port from `ClientHealthView`);
  - wrap the existing "Add client" form in the collapsed "Advanced enrolment options" pane.
- `server/frontend/src/lib/views/ClientHealthView.svelte` — delete once its pieces are ported.
- API wrappers in `server/frontend/src/lib/api/` (`clients.ts`, `client-health.ts`) are unchanged — same endpoints, just one consuming view.
- No server/API/DB changes required. This is a frontend consolidation only.
- Tests/smoke: update any view-level smoke coverage; ensure the admin nav no longer references `client-health`. (Note existing smoke-coverage patterns, cf. #285.)

## Acceptance criteria

- [ ] Admin nav shows a single **Clients** item; no **Client Health** item.
- [ ] The Clients view shows inventory + health together per client, with edit/delete and expandable queue detail.
- [ ] The primary "Enrol a client" action mints a token and shows the install one-liner (the former Client Health flow), with no endpoint changes.
- [ ] The manual `POST /api/clients` form lives in a collapsed-by-default "Advanced enrolment options" pane that explains the degraded (no bearer token / no links) trade-off.
- [ ] Manually-created clients are visually distinguishable from fully-enrolled ones in the list.
- [ ] `ClientHealthView.svelte` removed; quality gate (format/lint/typecheck/test) green.
- [ ] Docs/screenshots refreshed if the nav/view change is user-visible (cf. the `update-screenshots` flow).

---

_Filed from a UX-redundancy review of the two views; the underlying enrolment endpoints are sound, the issue is purely how the two views present overlapping entry points._

BenSeymourODB:
Closes #305.

## What

Consolidates the two overlapping admin surfaces — **Clients** (inventory CRUD) and **Client Health** (status board) — into one **Clients** view. Both previously offered to "enrol a client" via different, inconsistent mechanisms; now there is one home for clients and one canonical way to enrol them.

## Changes

- **Merged view** (`ClientsView.svelte`): joins the client inventory (`listClients`) with health (`listClientHealth`) keyed by id. Per client it shows hostname, SSH user, reachability, per-component health, and the offline queue, with inline edit/delete and the expandable queued-actions detail. Health degrades to `unknown` (and "awaiting first health probe") when no probe has run or the health fetch fails — inventory still renders.
- **Enrolment as the primary action**: the token-mint flow (mint a scoped, single-use token → render the documented `install-client.sh` one-liner) is promoted to the on-screen way to add a client. No endpoint changes.
- **Removed the separate surface**: dropped the `client-health` nav item, the `ClientHealthView` route branch, the component, and its test.
- **Removed the manual `POST /api/clients` create from the UI** (see rationale below). It remains an API-only capability for scripts/tests/integrations; the wrapper and its API-layer test are untouched.
- **New read-only `enrolled` boolean on `ClientResponse`** (true once a client holds a bearer token), so the merged view can flag a manual record as "not enrolled".

## Why the manual create was dropped from the UI (deviation from the filed plan)

The issue originally proposed keeping the manual `POST /api/clients` create behind a collapsed "Advanced enrolment options" pane. On closer inspection of the enrolment workflow that's the wrong call, because the manual create is not part of enrolment and is a foot-gun on the admin surface:

- `POST /api/clients/enrol` **always inserts** a fresh client row (`consumeTokenAndEnrol`) — it never claims/updates an existing one. With `hostname` uniquely indexed, a manual record for `mint-01` makes a later real enrolment of `mint-01` fail with a 409. So manual-create is *not* pre-registration; it collides with proper onboarding.
- A manual record carries **no bearer token** (no event stream / notifications) and **no supervised-user links**, and the dashboard's SSH key is never authorized on the box. It's inert until the admin separately authorizes SSH trust out-of-band and adds links via the User↔Client links view.

Its only genuine use is a narrow expert path (out-of-band provisioning, inventory placeholder, DB-row recovery), all needing manual follow-up and yielding a permanently degraded client. Surfacing that behind a disclosure triangle invites confusion, so it's kept as an API-only escape hatch instead. (Decision confirmed with the maintainer before pushing.)

## Tests / gate

- Server: `typecheck`, `lint`, `format:check`, `test` (1400) all green; coverage gate satisfied. Added `ClientResponse.enrolled` coverage to `policy-dtos.test.ts`.
- Frontend: `svelte-check` clean; full suite green. Split the merged view's coverage into `clients-view-crud.test.ts` (inventory edit/delete + the not-enrolled badge) and `clients-view-health.test.ts` (health list, queue, degraded-health fallback, and the enrol-token flow).

No docs/screenshots reference the removed view (none exist for it).

https://claude.ai/code/session_01V3ReAc946CcX9QWSS1w9qH

---
_Generated by [Claude Code](https://claude.ai/code/session_01V3ReAc946CcX9QWSS1w9qH)_


BenSeymourODB:
## Review: PR #312 — Merge "Client Health" into a single "Clients" view

This PR consolidates the **Clients** (inventory CRUD) and **Client Health** (status board) admin surfaces into one `ClientsView.svelte`, promotes the enrolment-token flow as the primary add-a-client action, drops the manual `POST /api/clients` create from the UI, and adds a read-only `enrolled` boolean to `ClientResponse`. The server-side DTO change is clean and well-tested (`enrolled = row.bearerTokenHash !== null`), the inventory↔health join is correct, and `load()` already fetches both endpoints in parallel. Five findings survived verification, most-severe first.

### Findings

1. **`server/frontend/src/lib/views/ClientsView.svelte:350` — "enrolled {date}" is shown even for records flagged "not enrolled", contradicting the new badge.**
   `enrolledAt` is non-nullable (`dtos.ts:120`) and `toClientResponse` always populates it from the row's creation timestamp (`dtos.ts:149`). A manual record has `enrolled: false` (the amber "manual · not enrolled" badge at line 343) yet a non-null `enrolledAt`, so the same card reads `manual · not enrolled` and, one line below, `enrolled 26 Jun 2026, 14:52`. That is exactly the confusion the `enrolled` flag was added (#305) to remove. The "enrolled <date>" line should be suppressed or relabeled (e.g. "added") when `!client.enrolled`.

2. **`server/frontend/src/lib/views/ClientsView.svelte:106` — a health-fetch failure is silently swallowed, hiding a broken health endpoint.**
   `listClientHealth().catch(() => [])` means any failure of `GET /api/clients/health` (auth expiry, 500, network) collapses every card to "unknown" reachability / "Awaiting first health probe" with no error shown. The removed `ClientHealthView` surfaced a `role="alert"` banner in this case. The admin can no longer distinguish "health service is down" from "clients genuinely un-probed." The degradation is deliberate per the comment, but consider surfacing a non-blocking notice (e.g. a "health unavailable" pill) so the signal isn't lost.

3. **`scripts/screenshots/capture.mjs:170` (also 39–40, 55–56, 179) — the screenshot generator still drives the deleted "Client Health" view.**
   The PR removed the `client-health` nav item, but `capture.mjs` still slugifies/captures `admin-client-health` and `admin-client-health-enrol`, and lists them in `ROOT_README_HIGHLIGHTS`. Running the documented `/update-screenshots` flow after this PR will never reach those screens (the nav label is gone), so those shots are skipped and the README-highlights check warns. The PR body's claim "No docs/screenshots reference the removed view" overlooked the generator; these slugs/captions should be folded into the merged `admin-clients` entry.

4. **`server/frontend/src/lib/views/ClientsView.svelte:155-157` — dead re-sync of `health[].hostname`.**
   `saveEdit` rewrites `health = health.map(... hostname: updated.hostname ...)`, but no part of the template renders `h.hostname` (the card shows `client.hostname` at line 342). The map allocates a new array on every rename for a field that is never read, and the comment ("Keep the joined health card's hostname in step") misleads a future maintainer into thinking the health card displays it. Delete lines 154–157.

5. **`server/frontend/tests/components/links-view-time-today.test.ts:49` — fixture omits the now-required `enrolled` field, masked by an `as ClientResponse` cast.**
   `clientResponseSchema` now requires `enrolled`, and the sibling `links-view.test.ts` was updated to add `enrolled: true`, but this fixture (`{ id: 7, hostname: "mint-01", ...overrides } as ClientResponse`) was not. The `as` cast suppresses the strict-mode error, so the test exercises a shape the real serializer never emits — diverging from production silently. Add `enrolled` to keep it consistent with the updated sibling.

Lower-severity (not surfaced as numbered findings): `expanded[]` (line 131) is never pruned in `handleDelete`, a small unbounded-growth/stale-state leak with no correctness impact; and `messageOf`/`formatDateTime` perpetuate pre-existing per-view duplication rather than introducing it.

PR #324, opened 2026-06-27
Serve install-client.sh from dashboard at GET /install-client.sh

BenSeymourODB:
## Summary

The script existed in the repo and was attached to GitHub Releases, but was never bundled into the Docker image or exposed via an HTTP route — so `curl https://<dashboard>/install-client.sh` returned 404 from a deployed alpha instance.

- **Widen Docker build context** from `server/` to the repo root; prefix all `COPY` paths with `server/` and copy `client/install-client.sh` into the image at `./client-scripts/install-client.sh`
- **Update CI/release workflows** to pass `--file server/Dockerfile` with `.` as the build context (`ci.yml` docker-build job; `release.yml` build-and-push step)
- **Add `installClientScriptPath` to `Settings`** (`PCT_INSTALL_CLIENT_SCRIPT_PATH`, default `/app/client-scripts/install-client.sh`) so dev and tests can override it
- **New `web/install-script.ts`**: registers `GET /install-client.sh`, streams the script with `Content-Type: text/x-shellscript`, warns at startup (and 404s at request time) when the file is absent — same graceful-degradation pattern as `web/frontend.ts`
- **Wire `registerInstallScript`** into `buildApp` alongside `registerFrontend`
- **New `tests/web/install-script.test.ts`**: 5 tests covering script-present (200, correct content type, other routes unaffected) and script-absent (404, startup warning, other routes unaffected) cases

After this lands, a client device can bootstrap with:

```bash
sudo bash <(curl -fsSL https://<dashboard>/install-client.sh) \
    --server-url https://<dashboard> \
    --enrolment-token <token> \
    --supervised-user <username>
```

## Test plan

- [ ] `npm run typecheck` passes
- [ ] `npm test` passes (5 new tests in `install-script.test.ts`, all green)
- [ ] Docker image builds with the widened context (`docker buildx build --file server/Dockerfile .`)
- [ ] `GET /install-client.sh` on a deployed container returns 200 with the script content

---
_Generated by [Claude Code](https://claude.ai/code/session_01Maq6PcuFdhRPFzDy2wq1uT)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,10 +197,11 @@ jobs:
         run: |
           if [ -f server/Dockerfile ]; then
             docker buildx build \
+              --file server/Dockerfile \
               --cache-from type=gha \
               --cache-to type=gha,mode=max \
               --tag pct-dashboard:ci \
-              server/
+              .
           else
             echo "server/Dockerfile not yet present — skipping Docker build"
           fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,7 +69,8 @@ jobs:
       - name: Build and push image
         uses: docker/build-push-action@v6
         with:
-          context: server
+          context: .
+          file: server/Dockerfile
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -10,8 +10,8 @@
 # docs/server-deployment.md and docs/licensing-analysis.md, enforced by
 # .github/workflows/license-guard.yml.
 #
-# Build context is server/ (see .github/workflows/ci.yml → docker-build),
-# so server/frontend/ is in scope for the frontend build stage.
+# Build context is the repo root (see .github/workflows/ci.yml → docker-build),
+# so both server/ and client/ are in scope.
 
 # ---------------------------------------------------------------------------
 # Stage 1: build the SvelteKit frontend (/admin + /app) to static assets.
@@ -20,10 +20,10 @@
 FROM node:22-slim AS frontend-builder
 WORKDIR /app/frontend
 
-COPY frontend/package.json frontend/package-lock.json ./
+COPY server/frontend/package.json server/frontend/package-lock.json ./
 RUN npm ci
 
-COPY frontend/ ./
+COPY server/frontend/ ./
 RUN npm run build
 
 # ---------------------------------------------------------------------------
@@ -40,11 +40,11 @@ RUN apt-get update \
   && apt-get install -y --no-install-recommends python3 make g++ \
   && rm -rf /var/lib/apt/lists/*
 
-COPY package.json package-lock.json ./
+COPY server/package.json server/package-lock.json ./
 RUN npm ci
 
-COPY tsconfig.json tsconfig.build.json ./
-COPY src/ ./src/
+COPY server/tsconfig.json server/tsconfig.build.json ./
+COPY server/src/ ./src/
 RUN npm run build
 
 # Strip dev dependencies in place, leaving the already-compiled native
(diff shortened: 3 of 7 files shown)


BenSeymourODB (commented):
## Code review

The approach is solid — widening the build context, updating both CI and release workflows, and following the graceful-degradation pattern from `web/frontend.ts` are all the right calls. Four findings below.

---

### 1. Unhandled stream error after response headers are committed

**`server/src/web/install-script.ts` — request handler**

```typescript
return reply.type("text/x-shellscript").send(createReadStream(scriptPath));
```

`reply.type(...).send(stream)` commits the `200` status and content-type header before the stream is consumed. If the `ReadStream` then emits an error (file deleted or permission change between the `existsSync` check and the read, or a filesystem hiccup), Fastify receives the error after headers may already be flushed — the client gets a partial `200` response that silently truncates instead of a clean error code.

The simplest fix is to attach an error listener that sends a 500 when the reply hasn't been committed yet:

```typescript
app.get("/install-client.sh", async (_request, reply) => {
  if (!existsSync(scriptPath)) {
    return reply.code(404).type("text/plain").send("install-client.sh not found");
  }
  const stream = createReadStream(scriptPath);
  stream.on("error", (err) => {
    if (!reply.sent) {
      reply.code(500).type("text/plain").send("Failed to read install-client.sh");
    } else {
      app.log.error(err, "install-client.sh stream error after headers sent");
    }
  });
  return reply.type("text/x-shellscript").send(stream);
});
```

---

### 2. Temp directory leaked in `afterEach` (script-present suite)

**`server/tests/web/install-script.test.ts` — `afterEach`**

```typescript
beforeEach(() => {
  const dir = mkdtempSync(join(tmpdir(), "pct-install-script-"));
  scriptPath = join(dir, "install-client.sh");
  writeFileSync(scriptPath, SCRIPT_CONTENT);
  // `dir` goes out of scope here — never stored
});

afterEach(async () => {
  await app.close();
  db.$client.close();
  rmSync(scriptPath, { force: true }); // removes the file, not the directory
});
```

`dir` is a local variable that's lost after `beforeEach` returns. Each test run leaks a directory under `tmpdir()`. Fix: store `dir` at the describe scope alongside `scriptPath`, then clean up the whole tree in `afterEach`:

```typescript
let dir: string;

beforeEach(() => {
  dir = mkdtempSync(join(tmpdir(), "pct-install-script-"));
  scriptPath = join(dir, "install-client.sh");
  // ...
});

afterEach(async () => {
  await app.close();
  db.$client.close();
  rmSync(dir, { recursive: true, force: true });
});
```

---

### 3. `afterEach` in the absent-file suite will throw `TypeError` if `buildApp` fails

**`server/tests/web/install-script.test.ts` — absent-file describe block**

```typescript
describe("GET /install-client.sh (script absent)", () => {
  let app: FastifyInstance; // undefined until assigned inside an it()
  let db: TestDb;

  afterEach(async () => {
    await app.close();     // TypeError if app was never assigned
    db.$client.close();
  });

  it("404s and emits a startup warning", async () => {
    db = testDb();
    app = buildApp({ ... }); // assigned here
    // ...
  });
```

Unlike the present-file suite there's no `beforeEach`, so `app` and `db` stay `undefined` until inside each `it`. If `buildApp` or `testDb` throws, `afterEach` runs against `undefined` and throws `TypeError: Cannot read properties of undefined (reading 'close')`, masking the original failure. Moving setup into a `beforeEach` (matching the present-file suite's structure) or guarding with `await app?.close()` / `db?.$client.close()` fixes this.

---

### 4. Per-request `existsSync` is a synchronous filesystem call on every download (low severity)

**`server/src/web/install-script.ts` — request handler, line ~39**

The file is copied into the image at build time and will never appear or disappear in production. The startup check at registration time already captures its presence. The per-request `existsSync` re-checks on every `GET` — a blocking syscall that stalls the event loop for the duration. It's a low-traffic endpoint so the practical impact is negligible today, but caching the boolean at registration time is simpler and removes both the redundancy and the blocking:

```typescript
const scriptPresent = existsSync(scriptPath);
if (!scriptPresent) { /* warn */ }

app.get("/install-client.sh", async (_request, reply) => {
  if (!scriptPresent) {
    return reply.code(404).type("text/plain").send("install-client.sh not found");
  }
  // ... stream
});
```

(If the self-healing "serve once the file appears after startup" behaviour is intentional for dev workflows, a comment explaining that would help.)

---

Overall this is close — findings 1–3 are worth addressing before merge.


---

Issue #81, opened 2026-06-16
Client health/status API + admin "Clients" page


BenSeymourODB:
## What

Surface enrolled machines and their per-component health in the admin UI, plus the `/api/*` endpoint that backs it. Mocked in PR #60 — [`design/admin/clients.html`](../blob/main/design/admin/clients.html).

## Scope

**Backend**
- A status/probe path that reports, per `Client`: reachability (SSH probe), `last_seen`, and component health for Timekpr-nExT / ActivityWatch / e2guardian / `pct-client-bridge` / `pct-client-agent`.
- Surface **offline + queued-change** state (ties into the Phase 4 offline-queue) so the admin sees what's pending for an unreachable client.
- zod DTOs in `server/src/api/`.

**Frontend (`/admin`)**
- Clients list with component-health badges, offline/queued indicators, and the **enrol-a-client** flow that mints an enrolment token (consumes the enrolment-token endpoint).

## Notes

- Maps to Phase 3 and the "Failure modes the design must handle" section of [`docs/architecture.md`](../blob/main/docs/architecture.md).
- Health probing should reuse the Phase 4 SSH transport facade rather than opening its own connections.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Scoping note: #81 has a **backend** half (client health/status API: SSH reachability probe, `last_seen`, per-component health, offline + queued-change surfacing, zod DTOs) and a **frontend** half (admin "Clients" page + enrol flow). The admin page is blocked on the in-flight `/admin/*` shell (#53), so this session lands the **backend health/status API end-to-end** — reusing the existing Phase-4 SSH transport facade and offline-queue, with zod DTOs and unit tests against injected SSH/probe seams. The admin "Clients" page UI will be split into a tracked follow-up issue and linked from the PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

BenSeymourODB:
Issue-hygiene flag (automated audit): the backend half of this issue has landed (PR #196 — client health/status API + SSH component prober), and the **frontend half was explicitly split out into #194** ("the frontend half of #81 … Clients health/status page + enrol-a-client flow"). The basic Clients CRUD editor also shipped via #189.

[...]

Issue #78, opened 2026-06-16
Client: pct-agent service-user provisioning + scoped sudoers


BenSeymourODB:
## What

Provision the dedicated `pct-agent` service account on the client and the **narrow** sudoers rules it needs, as described in [`docs/client-install.md`](../blob/main/docs/client-install.md). This is the SSH principal the dashboard's transport facade (Phase 4) connects as.

## Scope

- Create the `pct-agent` system user; install the dashboard's SSH public key into its `authorized_keys`.
- Scoped `sudoers` drop-in granting **only** the specific commands the toolkit invokes (notably `timekpra ...`, and later the Ansible-applied actions / `timekpra --kill-session` used by the bridge), never blanket root.
- Idempotent: re-running the installer reconciles rather than duplicates.

## Notes

- Phase 3 bullet "`pct-agent` user provisioning + scoped sudoers" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Keep the sudoers surface as tight as the license/security posture requires; the Phase 8b `pct-client-bridge` holds its own narrow rule (`timekpra --kill-session`, lockout clear/set) — coordinate the command list so it stays auditable.
- Respect the bounded tamper-resistance posture — this is least-privilege, not lockdown.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Picked as the first eligible roadmap item: Phase 2 is currently blocked behind open draft PRs (#102 for #52, and the #51 CRUD chain), and #76/#77 depend on the (blocked) enrolment endpoint + auth. #78 has no open blockers and is a self-contained client-side shell deliverable.

Plan: implement a sourceable, idempotent `pct-agent` provisioning step (system user + visudo-validated, least-privilege NOPASSWD sudoers drop-in for `timekpra`, optional authorized-key install), with shellcheck + bats tests. The Ansible-applied privileged sudoers entries stay deferred to their Phase-6 playbooks (#90/#92) with a documented extension point — keeping the granted surface as tight as the bounded tamper-resistance posture requires.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #341, opened 2026-06-29
feat: install+enable client SSH daemon and wire the live health prober

BenSeymourODB:
## Summary

Closes the two gaps that left the dashboard unable to reach freshly-enrolled clients (the `SSH probing not yet configured (#39)` / `unknown` state on the Clients page, and queued actions stuck at `pending ×0`).

### 1. Client install: install + enable the SSH daemon

`provision-agent-user.sh` set up the `pct-agent` SSH *principal* (account, scoped `timekpra`-only sudoers, `~/.ssh/authorized_keys`) but nothing ever installed or enabled an SSH *daemon*. Fresh Linux Mint ships no `openssh-server`, so there was nothing for the dashboard to connect to — which is exactly the "sshd service not active" error seen during setup, and why the offline-queue drainer skipped those clients (its reachability probe returns offline, so actions never even attempt → `×0`).

- `install-baseline-tools.sh` now installs `openssh-server` and enables `ssh.service`.
- The unit name is overridable via `PCT_SSHD_SERVICE` (default `ssh.service`), named to match `self-test.sh`'s existing variable so the install step and the self-test stay in lockstep.

### 2. Dashboard: wire the live `SshClientProber` (#81)

The `SshClientProber` was fully implemented and unit-tested but **never injected** — `registerClientHealthRoutes` was called without a `prober`, so `/api/clients/health` always degraded to `unknown` with `SSH probing not yet configured (#39)`.

- `createPolicyPushTransport` (the one place SSH credentials + the pooled transport are already wired) now builds the prober over that **same pooled SSH transport** — specifically its *un-audited* `exec` surface, since a `systemctl is-active` liveness query is data, not an admin command worth flooding the audit log with.
- The prober is exposed on `PolicyPushTransport` and threaded through `buildApp` → `registerApi` → the health routes.
- When no SSH key exists (dev / CI / tests / pre-first-run-keygen) the prober is withheld and the routes keep degrading gracefully — unchanged behaviour.

## License boundaries

No boundary touched. The prober only `exec`s `systemctl` over the existing SSH subprocess facade (no in-process GPL linkage), and `openssh-server` is installed via `apt` on the *client* — never bundled into the dashboard image.

## Testing

- `npm run typecheck`, `npm run lint`, `npm run format:check` — clean.
- `npm test` — 1619 tests pass, coverage gate (80%) held. New `bootstrap.test.ts` case asserts the live transport exposes a prober that probes over the shared SSH pool, and the fallback exposes none.
- `bats client/tests/*.bats` — 129 pass, including 3 new cases (openssh-server installed, `ssh.service` enabled, `PCT_SSHD_SERVICE` override). `shellcheck` clean.

## Deferred work (already tracked)

Two per-user health components remain reported as `unknown` by design — the loopback `aw-server` probe (**#86**, Phase 5) and the per-user `pct-client-agent` `systemd --user` probe (**#103**, Phase 8b). These are pre-existing deferrals with their own issues, not introduced here; this PR wires the system-service probes (Timekpr-nExT, e2guardian, `pct-client-bridge`) that were already defined.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NVEYE68aFNye6rGFPNiyCt)_


server/src/transport/policy-push/bootstrap.ts
```diff
@@ -165,6 +177,13 @@ export function createPolicyPushTransport(
   const dispatcher = createPolicyPushDispatcher({ db, executor, log });
   const probe = sshReachabilityProbe(db, ssh, credentials);
 
+  // The client health prober (#81) shares this same pooled SSH transport — but
+  // its *un-audited* `exec` surface: a `systemctl is-active` liveness query is
+  // data, not an admin command, so it must not flood the audit log the way the
+  // `timekpra` pushes (through `auditing`) deliberately do. buildApp injects it
+  // into the /api/clients/health routes.
+  const prober = new SshClientProber(ssh, credentials);
```
BenSeymourODB, line 185: The intentional un-audited probe path is the load-bearing decision in this PR, and it is correct on the merits: `systemctl is-active` is a read-only liveness query, and the health page can probe a whole fleet on every load, so routing it through `auditing` would bury the `timekpra` admin actions under probe noise. No license boundary is crossed either — `ssh` is still the subprocess SSH facade, just the un-decorated one.

Two things worth doing so this stays correct:

1. The guarantee is untested. The new test in `bootstrap.test.ts` proves the prober *works*, but nothing asserts that probing leaves the audit log empty — which is the whole point of passing `ssh` here instead of `auditing`. A future refactor that "consistently" wires the prober over `auditing` (note `AuditingTransport.exec` at `transport/audit/transport.ts:142` does audit unchecked execs) would silently flood the audit log and no test would fail. Please add an assertion that after a `prober.probe(...)`, `listAuditEntries(db, ...)` returns zero rows (the test file already imports `listAuditEntries`).

2. Consider a one-line note in `docs/architecture.md` (or wherever the audit log's coverage is described) that health probes are deliberately excluded from the audit trail. Right now the only record of that decision is this code comment; an auditor reading "every command the dashboard issues is recorded" (the wording in `transport/audit/index.ts`) would not expect a silent carve-out.

---

server/tests/transport/policy-push/bootstrap.test.ts
```diff
@@ -127,6 +130,34 @@ describe("createPolicyPushTransport", () => {
...
+    expect(result.reachability).toBe("online");
+    // The system-service components probed via `systemctl is-active` are `ok`.
+    expect(result.components.some((c) => c.component === "timekpr-next" && c.status === "ok")).toBe(
+      true,
+    );
+
+    transport.dispose();
+    expect(ssh.disposed).toBe(1);
+  });
+
```
BenSeymourODB, line 168: This test asserts too little for the behaviour the PR is built around.

1. It does not assert the prober is un-audited. The comment at `bootstrap.ts:185` says the probe "must not flood the audit log"; please add `expect(listAuditEntries(db, {}).length).toBe(0)` (or equivalent) after the `probe(...)` call. Without it, the un-audited wiring — the single most consequential choice in this PR — has no regression guard.

2. The probe walk has two non-trivial branches that go untested here: the SSH-error path (`exec` rejects with an `SshError` → `reachability: "offline"` + all components `unknown`, `prober.ts:153`) and the `deferred` components (`activitywatch`, `pct-client-agent` → `unknown`, `prober.ts:127`). The fake's `ssh.exec` always returns `active`, so only the happy `online`/`ok` branch is exercised. The offline classification in particular is the behaviour the `clientHealth` fan-out bounds exist to protect, so it's worth a case where `ssh.exec` rejects with `new SshUnreachableError(...)` (already imported in this file) and asserting `result.reachability === "offline"`.

These can live in `transport/health/prober.test.ts` if you'd rather keep the bootstrap test focused on assembly — but at least the no-audit assertion belongs here, since it's a property of the assembly, not of the prober in isolation.

---

client/install-baseline-tools.sh
```diff
@@ -271,6 +283,21 @@ pct_baseline_install_packages() {
...
+# this client. The dashboard connects as the pct-agent principal (account +
+# authorized_keys provisioned by lib/provision-agent-user.sh, #78) to push
+# `timekpra` policy, run health probes (#81), and tunnel telemetry (#86). The
+# openssh-server PACKAGE is installed in pct_baseline_install_packages above;
+# this step only ensures the daemon is running and persists across reboots.
+# Fresh Linux Mint installs no SSH server by default, so without this both the
+# dashboard connection and self-test.sh's sshd check (#80) fail.
+pct_baseline_configure_sshd() {
+  pct_step "Enable the OpenSSH server (the dashboard connects over SSH)"
+  pct_run systemctl enable --now "$PCT_SSHD_SERVICE"
```
BenSeymourODB, line 298: Placement and posture, both minor but worth a moment's thought.

Placement: `provision-agent-user.sh` already owns "make this box reachable as the SSH principal" (it creates `pct-agent`, installs its `authorized_keys`, and scopes its sudoers). Enabling sshd is the other half of that same concern — without the daemon the key/account is inert. There's a reasonable argument that `enable --now ssh.service` belongs alongside the account provisioning rather than in the generic tool-installer. That said, the package install (`openssh-server`) clearly belongs in `pct_baseline_install_packages`, and keeping the enable next to it so `self-test.sh`'s sshd check passes within the baseline run is a defensible call. The header comment already cross-references #78; I'd just flag the split for the maintainer to confirm it's intentional rather than incidental. Not blocking.

Security posture (worth one line of docs, not a code change): `systemctl enable --now ssh.service` opens password-and-key SSH to the whole network on a child's desktop, persistent across reboots. The product's threat model intentionally relies on `pct-agent` being key-only + sudoers-scoped, but enabling the daemon also exposes every *other* local account (including the child's, and root if `PermitRootLogin` is left at the distro default) to network login. This is squarely within the documented "bounded tamper-resistance" posture — do not add hardening like disabling password auth or locking down `sshd_config` as anti-circumvention; that would be scope creep against the CLAUDE.md ceiling. But a sentence in `docs/client-install.md` noting that enabling sshd widens the host's network surface (and that locking it down further is deliberately out of scope) would set expectations honestly.

---

client/tests/install-baseline-tools.bats
```diff
@@ -147,6 +147,24 @@ plan() { # run the script in dry-run with the given args, capture the plan
...
+@test "enables the OpenSSH server so the dashboard can reach the client" {
+  plan --supervised-user alice
+  [ "$status" -eq 0 ]
+  [[ "$output" == *"systemctl enable --now ssh.service"* ]]
+}
+
+@test "the SSH daemon unit name is overridable (PCT_SSHD_SERVICE)" {
+  PCT_SSHD_SERVICE=sshd.service plan --supervised-user alice
+  [ "$status" -eq 0 ]
+  [[ "$output" == *"systemctl enable --now sshd.service"* ]]
```
BenSeymourODB, line 170: These three tests are good (package install, enable, and the `PCT_SSHD_SERVICE` override). One small gap: nothing asserts the sshd step is sequenced *before* the steps that depend on a reachable client, and nothing guards against the baseline silently skipping it. The package and enable asserts are independent substring matches on the same plan, so a refactor that dropped `pct_baseline_configure_sshd` from `pct_install_baseline_tools` (line ~564) but left the package in the install list would still pass two of the three. Optional, but an assertion that the enable line appears after the `apt-get install` line in `$output` would lock the ordering. Not blocking.

---

BenSeymourODB (commented):
## Review: #341 — sshd in baseline install + health prober over pooled SSH

Overall this is a clean, well-scoped change and the architecture is right. Verdict: **approve-with-nits** (one of which — the missing no-audit test assertion — I'd like addressed before merge, but it's a test addition, not a redesign).

### License boundaries — clean
The prober only `exec`s `systemctl is-active` over the existing SSH subprocess facade. No GPL code is linked in-process, no subprocess/REST boundary is collapsed, and Timekpr/Ansible/e2guardian/ActivityWatch integration is untouched. The catalogue (`components.ts`) only *names* the units to probe. No new `any`, no `as` casts, no `@ts-ignore`; the optional-field plumbing uses the project's established conditional-spread idiom for `exactOptionalPropertyTypes`. `tsc`/lint posture is preserved.

### The one decision that matters: un-audited probes
Passing the raw `ssh` transport (not the `auditing` decorator) into `SshClientProber` is **correct** — a fleet-wide liveness query on every Clients-page load should not bury `timekpra` admin actions in the audit log. But the guarantee is undocumented outside a code comment and, more importantly, **untested**: nothing asserts probing leaves the audit log empty, so a future "let's wire it through `auditing` for consistency" refactor would silently regress it with a green suite. See the inline comment on `bootstrap.ts:185` and the test comment on `bootstrap.test.ts`.

### Tamper-resistance scope
Within bounds. Enabling sshd widens the client's network surface; the right response is a docs sentence, **not** hardening `sshd_config` — I've explicitly flagged that adding anti-circumvention lockdown would breach the CLAUDE.md ceiling.

### Summary of asks
- **Should-fix (test):** assert no audit entries are written by a probe (inline on `bootstrap.test.ts`).
- **Nice-to-have (test):** cover the offline (`SshError`) and `deferred`-component branches of the probe walk.
- **Nice-to-have (docs):** note in `docs/client-install.md` that enabling sshd widens the host's network surface and that further lockdown is out of scope; note in the audit docs that health probes are deliberately excluded from the trail.
- **Confirm (placement):** intentional that the sshd *enable* lives in `install-baseline-tools.sh` rather than `provision-agent-user.sh` (which owns the rest of "reachable as pct-agent").

None of these block the core change; the client install correctly installs+enables `openssh-server` (matching `self-test.sh`'s `PCT_SSHD_SERVICE` default and dry-run aware), and the prober is cleanly threaded through `app.ts` → `plugin.ts` → the health routes with a faithful pre-#39 degrade-to-`unknown` fallback.

---

BenSeymourODB:
Thanks for the review — pushed in 3a6b105. The should-fix (no-audit regression guard) and the nice-to-have docs/test items are addressed; the offline (`SshError`) and `deferred`-component probe branches are already covered in `transport/health/prober.test.ts`, so I didn't duplicate them.

On the one item that's a decision rather than a change — **sshd placement** (`install-baseline-tools.sh` vs `provision-agent-user.sh`): keeping it in the baseline installer is intentional. `provision-agent-user.sh` is deliberately a pure account/file module — it sources *before* `pct-common.sh`, sets no shell options at file scope, never `exit`s, and runs unprivileged in its bats suite, so it has none of the `apt`/`systemctl`/`pct_run`/dry-run machinery that installing+enabling a daemon needs. The package install clearly belongs in `pct_baseline_install_packages`, and keeping the `enable --now` next to it means `self-test.sh`'s sshd check passes within a single baseline run. The header already cross-references #78. Happy to revisit if you'd rather split it, but that's the reasoning for the current split.

(Separately: the bats CI failure was a host-dependent assertion — GitHub runners already ship `openssh-server`, so the dry-run reported it already-installed instead of in the `apt-get` argv. Fixed to accept either branch.)

---

PR #344, opened 2026-06-29
feat(client): support upgrade-in-place re-runs of the installer

BenSeymourODB:
## Summary

Follow-up to #341. Deploying that sshd-enable fix to the already-enrolled clients means re-running the installer — but a plain re-run **aborted at the enrol step**: the enrolment token is single-use (`401`) and the hostname is unique on the dashboard (`409`), and under `set -euo pipefail` either one killed the run. This makes re-running an already-enrolled client a first-class, supported path.

## Changes

**Two re-run shapes in `install-client.sh`:**

1. **`--skip-enrol`** (and `PCT_SKIP_ENROL=1`) — re-runs provision + baseline + self-test only, skipping the enrolment exchange and its dependents (authorizing the SSH key, persisting the per-client token). Needs no token and contacts no dashboard, so an existing enrolment / token / authorized key is left untouched. The `--enrolment-token` and `--server-url` requirements are relaxed accordingly.

   ```bash
   sudo bash install-client.sh --skip-enrol --supervised-user alice
   ```

2. **409-tolerant plain re-run** — `pct_orch_enrol` now captures the HTTP status (via `--output`/`--write-out` instead of `--fail`) and returns a distinct exit code for the dashboard's "already enrolled" `409`. The orchestrator treats that as a tolerable upgrade condition: it keeps the existing enrolment and continues to the self-test instead of aborting. A genuinely bad/expired token still `401`s and aborts with a clear message. (The baseline step runs before enrol, so the client-side patch lands either way.)

**Docs:** new "Upgrading an already-enrolled client" section in `docs/client-install.md` covering both shapes, plus the two things a client-side re-run does **not** do and that must be handled separately — redeploy the dashboard (the served bundle is baked at image-build time, and server-side changes like the live health prober only take effect on redeploy), and re-authorize the dashboard's current public key (only a real enrol writes `authorized_keys`).

## License boundaries

None touched — pure bash orchestration; GPL tools still come from the distro/upstream via the baseline step, nothing linked in-process or bundled into the image. No tamper-resistance scope creep (this is about idempotent re-runs, not lockdown).

## Testing

- `shellcheck client/install-client.sh` — clean.
- `bats client/tests/*.bats` — 133 pass, including 4 new/updated cases: 409 tolerance, `--skip-enrol`, `PCT_SKIP_ENROL=1`, and the token-on-stdin stub updated for the new curl invocation.
- No `server/` files changed, so the TS hooks/CI are unaffected; file-hygiene (trailing whitespace, final newline) verified.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NVEYE68aFNye6rGFPNiyCt)_

### DIFF
--- a/client/install-baseline-tools.sh
+++ b/client/install-baseline-tools.sh
@@ -140,6 +140,36 @@ E2G_PCT_DIR="${E2G_PCT_DIR:-${E2G_DIR}/pct.d}"
 # named to match self-test.sh's PCT_SSHD_SERVICE so the two stay in lockstep.
 PCT_SSHD_SERVICE="${PCT_SSHD_SERVICE:-ssh.service}"
 
+# The systemctl binary, overridable so the bats suite can point it at a recorder
+# stub and assert which units get (re)started without a live systemd.
+PCT_SYSTEMCTL="${PCT_SYSTEMCTL:-systemctl}"
+
+# Restart/reload a system unit to APPLY a config change, but only when the config
+# actually changed across this run. `before` is a pct_file_checksum captured
+# before the write; if the file's checksum still matches, this is a no-op — so a
+# re-run that changes nothing never bounces the daemon. Without this, the baseline
+# only `enable --now`s a unit (a no-op for an already-running one), so on a re-run
+# changed config is written to disk but the running daemon keeps the old config
+# until a manual restart — the upgrade-in-place gap this closes (#347).
+#
+# `action` is a systemctl verb; use the `try-*` forms (try-restart,
+# try-reload-or-restart) so an inactive unit is a clean no-op rather than an
+# error under the caller's `set -e`. Dry-run prints the conditional intent.
+pct_apply_change() {
+  local file="$1" before="$2" unit="$3" action="$4"
+  if pct_is_dry_run; then
+    printf '%s %s %s %s (only if %s changed)\n' \
+      "$PCT_DRYRUN_PREFIX" "$PCT_SYSTEMCTL" "$action" "$unit" "$file" >&2
+    return 0
+  fi
+  if [ "$before" = "$(pct_file_checksum "$file")" ]; then
+    pct_log "${file} unchanged; not restarting ${unit}"
+    return 0
+  fi
+  pct_log "${file} changed; ${action} ${unit} to apply"
+  "$PCT_SYSTEMCTL" "$action" "$unit"
+}
+
 # --- step: apt repositories ------------------------------------------------
 
 # Resolve the Ubuntu series (apt "Suite") the PPA is consumed for. Mint reports
@@ -295,7 +325,7 @@ pct_baseline_install_packages() {
 # dashboard connection and self-test.sh's sshd check (#80) fail.
 pct_baseline_configure_sshd() {
   pct_step "Enable the OpenSSH server (the dashboard connects over SSH)"
-  pct_run systemctl enable --now "$PCT_SSHD_SERVICE"
+  pct_run "$PCT_SYSTEMCTL" enable --now "$PCT_SSHD_SERVICE"
 }
 
 # --- step: ActivityWatch upstream bundle -----------------------------------
@@ -347,10 +377,15 @@ pct_baseline_install_activitywatch() {
 
 pct_baseline_configure_timekpr() {
   pct_step "Configure Timekpr-nExT (baseline: daemon + generous Alpha-1 warnings)"
+  # Snapshot the config before editing so we only restart the daemon if the
+  # edit actually changes it (a no-op re-run must not bounce it).
+  local conf_before
+  conf_before="$(pct_file_checksum "$PCT_TIMEKPR_CONF")"
+
   # Initial policy is intentionally empty — the dashboard pushes limits via
   # `timekpra` over SSH after enrolment. We only ensure the daemon is up and
   # the CLI the server drives is on PATH.
-  pct_run systemctl enable --now timekpr.service
+  pct_run "$PCT_SYSTEMCTL" enable --now timekpr.service
 
   # Give supervised users plenty of advance warning before a session cutoff.
   # Alpha-1 has no pct-client-agent cadence/grace UX (Phase 8b), so this is the
@@ -360,6 +395,13 @@ pct_baseline_configure_timekpr() {
     TIMEKPR_FINAL_NOTIFICATION_TIME "$PCT_TIMEKPR_FINAL_NOTIFICATION_TIME"
   pct_set_conf_key "$PCT_TIMEKPR_CONF" \
     TIMEKPR_FINAL_WARNING_TIME "$PCT_TIMEKPR_FINAL_WARNING_TIME"
+
+  # `enable --now` started the daemon with whatever timekpr.conf was on disk
+  # then; if the edits above changed it (e.g. on an upgrade re-run), restart so
+  # the new warning lead times take effect. Timekpr persists usage, so a restart
+  # does not reset today's budgets. `try-restart` is a no-op if the unit is not
+  # active, so it never errors under `set -e`.
+  pct_apply_change "$PCT_TIMEKPR_CONF" "$conf_before" timekpr.service try-restart
 
   if pct_is_dry_run; then
     printf '%s %s\n' "$PCT_DRYRUN_PREFIX" "command -v timekpra" >&2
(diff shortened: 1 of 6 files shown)


BenSeymourODB (commented):
## Fresh-eyes review of #344

I traced the diff against the surrounding code, ran the bats suites, and verified the systemctl/curl behaviour and the timekpr/e2guardian assumptions against upstream. Overall the two features are well-built and the comments are unusually thorough. The `pct_orch_enrol` rewrite, the `response=... || status=$?` exit-code capture, the temp-file lifecycle, and the token-on-stdin handling are all correct. The license-boundary and tamper-resistance guarantees hold (pure apt/systemd orchestration; nothing linked or bundled; nothing anti-circumvention). My one material finding is a test-coverage gap, plus a couple of low-severity assumptions worth a sentence in the code. Verdict: **approve-with-nits**.

---

### Findings, most-significant first

**1. CONFIRMED — the `--skip-enrol` validation-relaxation is not actually pinned by any test (`client/tests/install-client.bats:368-391`)**

The two new orchestrator-level tests (`with --skip-enrol, re-runs ...` and `PCT_SKIP_ENROL=1 is equivalent ...`) run via `run env bash "$SCRIPT" ...`, and `setup()` (line 22) exports `PCT_DRY_RUN=1` into that environment. So both tests execute in **dry-run mode**. Under dry-run, `pct_orch_main` fills placeholders for a missing `--server-url`/`--enrolment-token` *before* the skip-enrol relaxation branch is ever reached (the `if pct_is_dry_run` arm at lines 654-656 / 665-668 wins). I verified this empirically: the run prints `no --server-url given; using placeholder ... for the dry-run preview` — the dry-run arm, not the `elif pct_orch_skip_enrol` arm.

Proof it asserts too little: I reverted the relaxation guard in place (`if [ -z "$enrol_token" ] && ! pct_orch_skip_enrol;` → `if [ -z "$enrol_token" ];`, i.e. re-broke the token requirement for skip-enrol) and **both tests still passed**. A regression that re-introduces the token/server-url requirement for a real `--skip-enrol` run would ship undetected.

Failure scenario: a future refactor drops the `pct_orch_skip_enrol` clause from the token check → real `sudo bash install-client.sh --skip-enrol --supervised-user alice` aborts with "missing --enrolment-token" → green CI. Fix: add a test that runs with `env -u PCT_DRY_RUN` (like `plan_strict` / the function-level enrol tests already do) and asserts a skip-enrol invocation with no token/server-url gets *past* validation (e.g. reaches the provision step) rather than exiting 2. The `[[ "$output" != *"/api/clients/enrol"* ]]` and `!= *"authorize dashboard ssh key"*` assertions are good and *do* pin that enrol/key-authorize are skipped — keep those; this is only about the validation relaxation.

**2. PLAUSIBLE (low) — `pct_apply_change` may issue a spurious timekpr restart because the `before` checksum is captured before `enable --now` (`client/install-baseline-tools.sh:382-388, 404`)**

`conf_before` is snapshotted at line 383, then line 388 does `enable --now timekpr.service`, then the two `pct_set_conf_key` edits, then the change-comparison at line 404. Timekpr-nExT's daemon is known to read and re-persist `/etc/timekpr/timekpr.conf` on startup (normalising/adding defaults). On the *first* run after a reboot where the unit was inactive, `enable --now` can rewrite the file independently of our two edits, so the checksum diff fires and we restart even when our keys were already correct. This is not a correctness bug — a restart is explicitly safe (usage is persisted, per your comment) — so impact is just an unnecessary daemon bounce, and on a steady-state no-op re-run (unit already active, `enable --now` a no-op) it behaves correctly. I could not find authoritative upstream docs confirming the rewrite-on-start, so flagging as PLAUSIBLE for your judgement. If it bothers you, snapshot `conf_before` *after* `enable --now` instead of before; that still catches the edits while ignoring the daemon's own normalisation.

**3. CONFIRMED (informational, behaviour is correct) — e2guardian has no `ExecReload`, so `try-reload-or-restart` will always restart, not reload (`client/install-baseline-tools.sh:604-605`)**

The stock e2guardian systemd unit ships no `ExecReload=` (upstream e2guardian#182). `systemctl try-reload-or-restart` correctly degrades to a restart when reload is unsupported, so the config change *is* applied — the code is right and the comment ("restarts if it doesn't [support reload]") is accurate. Noting only so nobody later "optimises" this to a plain `reload` expecting the ~1s path; on a stock box that would be a no-op-or-error, not a reload.

**4. PLAUSIBLE (low) — the e2guardian apply is keyed only on `e2guardianf1.conf`, so changes to `/etc/default/e2guardian` or the `pct.d/*.filtergroup` skeletons don't trigger anything (`client/install-baseline-tools.sh:573, 578-595, 604`)**

`pct_apply_change` only diffs `${E2G_DIR}/e2guardianf1.conf`. The function also writes `/etc/default/e2guardian` (line 578-579) and per-user skeletons (588-595). This is defensible: `/etc/default/e2guardian` is read only at service start (a reload wouldn't pick it up anyway, and on first install f1.conf goes absent→present so a restart happens regardless), and the `pct.d` skeletons are explicitly inert placeholders for Phase 6 Ansith, not read by the baseline. So no action needed today — but worth a one-line comment that the f1.conf checksum is a deliberate proxy for "baseline filter rules changed", lest a future edit to those other files be silently un-applied on a re-run.

**5. Minor — `pct_apply_change` propagates a real `try-*` failure under `set -e` (`client/install-baseline-tools.sh:165`)**

`"$PCT_SYSTEMCTL" "$action" "$unit"` is the function's last command, so a genuine restart failure (unit active but fails to come back) returns non-zero and aborts the caller. That's arguably the *right* behaviour (a failed daemon restart should surface), but it slightly contradicts the "never errors under set -e" framing in the comments, which is only true for the *inactive-unit* case. Consider tightening the comment.

---

### Things I checked and found clean (no action)

- **`pct_orch_enrol` rewrite:** body→`mktemp` (0600), `--output` keeps the body off stdout, `--write-out '%{http_code}'` yields a clean status with no trailing newline; `rm -f "$body_file"` is on all four paths (curl-fail / 2xx / 409 / other). Token still only on stdin via `--config -`, never argv. Dropping `--fail` is handled: without it curl exits 0 for any received HTTP response, so 409 lands in the `409)` arm and a connection error (non-zero exit, code `000`) is caught by the `if !` first. 3xx falls to `*)` → abort, which is correct (no `--location`).
- **`response=... || status=$?`:** `local ... status=0` is on its own line (line 528), so the exit code is not masked by `local`'s return; the 409 path correctly skips authorize-key/persist but still falls through to the self-test; a real failure returns 1 and aborts. Correct.
- **`pct_orch_skip_enrol` global consistency:** `PCT_SKIP_ENROL` is set from the local at line 644 *before* both the validation checks (657/665) and `pct_install_client` (686) read it via the predicate. No ordering hole.
- **Injection/quoting:** `PCT_SYSTEMCTL`, `$action`, `$unit` all passed as separate quoted words; no eval; bats stubs benign.
- **License boundary / tamper scope:** clean on both counts.
- **`sha256sum` availability:** present on the Mint/Ubuntu target and CI; acceptable.

bats: all 62 cases pass locally.

**Verdict: approve-with-nits.** The features are correct and safe to merge. Please address finding #1 (it's the only one that can hide a real regression) — ideally before merge; the rest are low-severity comment/robustness nits you can take or leave.

---

BenSeymourODB:
Thanks — addressed in `a290452`.

**Finding #1 (must-fix):** added real-run tests (`env -u PCT_DRY_RUN`) that stub `pct_install_client` and assert a `--skip-enrol` / `PCT_SKIP_ENROL=1` run with no token/server-url gets *past* validation, plus a negative control proving the guard still bites without skip-enrol. Verified by mutation: reverting the `&& ! pct_orch_skip_enrol` guard now fails the new test (it didn't before), so the relaxation is genuinely pinned.

**Nits taken:**
- 2: snapshot the timekpr conf checksum *after* `enable --now`, so the daemon's start-time normalisation of `timekpr.conf` can't read as our edit and cause a spurious restart.
- 4/3: comment that the e2guardian apply keys on `e2guardianf1.conf` as a deliberate proxy for "baseline rules changed," and that stock e2guardian ships no `ExecReload` so `try-reload-or-restart` is effectively a restart (change still applied).
- 5: clarified `pct_apply_change` only no-ops for an *inactive* unit — a genuine active-unit restart failure surfaces and aborts, by design.

Left as-is per your read.

---

Issue #343, opened 2026-06-29
Admin UI consolidation (pass 2): combined Budgets + Schedules + Group Schedules view with inline Policy Preview and effective-schedule view


BenSeymourODB:
## What

Second pass of the admin nav consolidation begun in #342 (which folded User
Groups into Users, Activity Groups into Activities, and moved "Add time today"
to the Dashboard). This pass collapses the per-user **policy editing** surfaces
into one coherent view.

Today these are four separate top-level nav sections:

- **Budgets**
- **Schedules**
- **Policy preview** (`PolicyPreviewView`)
- **Group Schedules**

They edit one conceptual thing — a user's (or group's) effective policy — but
are scattered, and the standalone Policy preview is effectively dead weight (see
below). Combine them into a single **Policy** view.

## Goals

1. **Combine Budgets + Schedules into one view.** A user's budgets and their
   schedule rules are the two halves of "what is this child allowed, and when" —
   edit them side by side for one selected user rather than in two disconnected
   sections.

2. **Use Policy Preview *inline* in the combined view.** The separate
   `PolicyPreviewView` is a standalone what-if surface, but the Budgets and
   Schedules editors **already submit their changes on save**, so by the time you
   could navigate to the preview the diff is already applied — the preview has
   nothing left to show. Hosting the preview bar **inside** the combined editor,
   reading the in-progress (unsaved) edits, restores its value: you see the
   push diff + affected clients *before* committing. This is exactly the
   "inline placement" bullet already scoped in **#304** (and the reusable
   preview bar shipped in #281 / PR #299); this issue is where it lands.

3. **Fold Group Schedules into the same view.** `GroupSchedulesView` becomes part
   of the combined surface, with:
   - **All active schedules displayed** — both individual *and* group rules — so
     the admin sees the full picture in one place.
   - **A combined target dropdown** for "add a schedule to…" that lists both
     users and user-groups (instead of today's separate user-schedule vs.
     group-schedule entry points).
   - **An effective-schedule view for an individual user** — resolve the
     user's own rules over their groups' rules and show the merged result
     (with inherited-vs-local marking), so the admin can answer "what will
     actually apply to Alice?" without mentally merging two lists.

## Relationship to existing issues

- **#304** — *Wire the save-and-push preview bar into the per-user policy
  editor + a "Save & push" action.* Directly in this path: its first bullet is
  the inline-preview placement above. This issue subsumes that placement work
[...]

Issue #182, opened 2026-06-19
Group-targeting on Schedule/Exception + user-over-group resolution (#124 follow-up)


BenSeymourODB:
## What

The second half of #124. The `UserGroup` entity + multi-group membership + `/api/user-groups` CRUD landed in **PR #181**; this issue covers letting policy *rules* target a group and resolving them with per-user override precedence.

- **Schema**: let `Schedule` and `Exception` target a `UserGroup` as well as a `User`. Today both are `user_id` NOT NULL. Options to weigh deliberately (this is the open design call):
  - relax `user_id` to nullable + add nullable `user_group_id` + an exactly-one-of CHECK (one table, but ripples through `ScheduleResponse`/`ExceptionResponse` — `userId` becomes nullable on the wire), **or**
  - a separate `group_schedules` / `group_exceptions` table (no change to the existing user-keyed tables, at the cost of duplicating the rule shape).
- **Resolution helper** `gatherUserScheduleRules(userId)` → the merged, precedence-ordered rule list: the user's **own** rules first (they win), then inherited group rules, preserving #63's first-match-wins *within* each level (ADR 0004), re-sequenced so it drops straight into `resolveEffectiveRule` (`policy/schedule-precedence.ts`). Tag each rule with its source (user vs which group) for the inherited-vs-local editor.
- **API**: extend the schedule/exception create DTOs/routes to accept a group target (mutually exclusive with the user target).
- **Tests** throughout.

## Why split out of #181

The roadmap deliberately builds the **effective-policy resolution engine** (#143, in-flight as **PR #176**) as the single "what applies for user U on day D" computation so downstream work "isn't coded against an interim contract." A user-over-group schedule resolver should compose *with* that engine, not duplicate it — so it should land **after #176 merges** and read through the same resolver. Building it now would couple it to an interim contract.

## Blocked by

- #176 (effective-policy resolver) — so the group layer plugs into the real resolver rather than an interim one.

## Also depends on / relates to

- Foundation: PR #181 (`UserGroup` + membership + CRUD). 
- The admin UI for inherited-vs-local (manage groups, assign users) stays in #124 and is blocked by #53/#63.
- Group-level **budgets** are the budget counterpart, tracked by #134.

## References

- `docs/architecture.md` → "Policy model"; `docs/adr/0004-schedule-precedence.md`; `docs/adr/0005-recurrence-and-date-scoping.md`.
[...]

Issue #142, opened 2026-06-17
Date-specific / future-dated rules (calendar overrides for specific days)


BenSeymourODB:
## What

Support **specific calendar dates in the future** — one-off, date-anchored rules/overrides such as "no screen time on 2026-06-30", "+2h on their birthday next month", or "screen-free vacation week (date range)". Today this is impossible: `exceptions` are active *immediately* until `expires_at` (no future start). An additive capability on the Phase 2 foundation, so it stays in Phase 13.

## Scope

- Use the `effective_from` / `effective_to` columns reserved in #146 (per ADR #139) so an override targets a future date or range. CRUD + zod DTOs.
- Compose cleanly with recurring schedules (#140) and budgets: define precedence (recommend date-specific overrides win over recurring rules for the days they cover) and how they layer with active grants — implemented as a layer in the resolution engine (#143).
- Resolution via #143 so a future-dated override surfaces in "coming up" (`/app`, client dashboard) and is pushed to the client when its window arrives (queueing if the client is offline, #84).

## Dependencies

- **Blocked by:** #139 (ADR), #146 (schema columns); **extends** #143 (resolution engine).
- These date-anchored rows are exactly the **dated, accumulating** data that motivates retention — cross-links #135 (purge past-dated overrides per policy).

Stays in Phase 13 (additive capability on the Phase 2/4 foundation).

BenSeymourODB:
Scope addition from the budgets/schedules/exceptions interaction review: when this issue's ADR defines how exceptions compose with recurring rules, it should also **pin down the `extend` schedule action**. `"extend"` has been in `scheduleActionValues` since the #60 mockups, but `resolveAllowedWindows` (`policy/resolve.ts:255`) treats it as a plain `allow` (`allowed = winner.action !== "deny"`) and nothing else distinguishes it — it's currently an undefined synonym.

The natural meaning: an `extend` rule/exception **widens** the allowed window past what the standing rules produce (rather than participating in first-match as an allow), which is exactly the mechanism **#364** needs for Nintendo-style paired "adjust bedtime as well" same-day extensions. If a different composition is chosen (or `extend` is dropped in favour of plain allow-exceptions winning over schedules), #364 should be designed against that decision — either way, decide it here in the ADR.

---
_Generated by [Claude Code](https://claude.ai/code)_

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

[...]

Issue #141, opened 2026-06-17
Day-of-week-varying time budgets (e.g. weekday vs weekend quotas)


BenSeymourODB:
## What

Let a time **budget vary by day of week** — e.g. "1h on school days, 3h on weekends" — which the current model cannot express (`budgets.window` is a single flat `daily|weekly|monthly` quota). An additive capability layered on the Phase 2 foundation, so it stays in Phase 13.

## Scope

- Represent weekday-varying allowances: a **budget profile** (per-weekday `seconds_allowed`) or per-weekday rows keyed to a profile. Migration + zod DTOs.
- Resolution: the effective daily quota for date D = the profile's value for D's weekday (in the user's effective TZ), before grants — plugs into the resolution engine (#143) as a composition layer, no rewrite.
- Wire into the grant recompute (#117) and the budget push (timekpra daily limit, #83).
- Admin editor + display (`/admin`, `/app`, client "My Time") showing the weekday pattern.

## Dependencies

- **Blocked by:** #139 (ADR), #146 (schema columns); **extends** #143 (resolution engine).
- Composes with group-level budgets (#134): a group defines the weekday profile, a user can override it.

Stays in Phase 13 (additive capability on the Phase 2/4 foundation).

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Scoping this run to the **backend slice** — schema + resolver + push + API CRUD + tests — which is implementable and fully unit-testable end-to-end now (blockers #139/#146 are closed; the resolver `#143`/push already reserve per-weekday seams that anticipate this, e.g. `policy-push/resolve.ts` replicates one daily value across all seven days pending #141, and `diff.ts` already emits per-weekday rows when days differ):

- **Schema:** a nullable 7-bit ISO-weekday `recurrence_days` mask on `budgets` and `group_budgets` (mirroring the reserved schedule/exception recurrence columns from #146/ADR 0005), constrained to `daily`-window budgets. Drizzle migration via `db:generate`.
- **Resolver:** a pure `selectBudgetsForWeekday` step inside `effectivePolicy` — per `(scope, window, target)` slot, a weekday-specific row (mask covers day *D*) wins over the uniform row; no covering row and no uniform → no daily limit that day. Covers overall **and** per-activity daily budgets. Weekly/monthly rolling budgets are unaffected (mask disallowed there).
[...]

PR #365, opened 2026-07-02
fix(policy): wire group-inherited schedules/budgets into enforcement (#362)

### LINKED ISSUE #362: Enforcement: group-inherited schedules/budgets never reach the client — wire gather* helpers into the policy push, force-close sweep, and preview
## Context

#182 and #134 landed the group layer end-to-end **except enforcement**: `group_schedules` / `group_exceptions` / `group_budgets` tables, full `/api` CRUD, and the user-over-group resolution helpers `gatherUserScheduleRules` / `gatherUserBudgets` (`server/src/policy/group-resolution.ts` — own rules first, then groups; dense re-sequenced ordinals; full-replace per budget slot; `RuleSource` tagging).

Today those helpers are consumed by exactly **one** caller: the display endpoint `GET /api/users/:userId/effective` (`server/src/api/policy/effective.ts:160–163`).

Every path that **actually enforces** reads the user's *own* rows only:

- **Policy push** — `transport/policy-push/executor.ts` (~56–57, 151–152) reads `listUserBudgets` + `listUserSchedules` before `resolvePolicyPush` → timekpra. A group-targeted bedtime/allowed-hours rule or group budget is **never pushed to any client**.
- **Per-activity force-close sweep** — `server/src/enforcement/evaluate.ts` reads schedules/budgets/grants by `userId` (own-only), so group per-activity quotas never trigger enforcement.
- **Save-and-push preview** — `api/policy/preview-routes.ts` (~13–15, 132) deliberately mirrors the push, so it also omits the group layer.
- **Ansible renders** — the e2guardian banned-sites render (`transport/ansible/e2guardian.ts`, `resolveBannedSites`) and the AppArmor hard-deny render (`transport/ansible/apparmor.ts`, `listUserAlwaysOnDenies`) also consider own-only always-on denies; an always-on group deny is ignored there too.

The result is a silent correctness gap: **group rules are display-only.** The group-mutation routes even fan a push out to every member's clients (`groupMemberPushCommands`, `api/policy/routes.ts:193–204`) — but the pushed payload is computed from own-only rows, so the push triggered by editing a group schedule doesn't contain the group schedule. `/admin`'s effective view and the actual client behaviour disagree.

## Scope

- Replace the own-only reads with the gather helpers (or an equivalent shared entry point) in:
  - [ ] `transport/policy-push/executor.ts` (timekpra limits + allowed-hours),
  - [ ] `enforcement/evaluate.ts` (per-activity quota decisions),
  - [ ] `api/policy/preview-routes.ts` (preview must keep mirroring what the push would do),
  - [ ] the e2guardian and AppArmor renders (inherited always-on group denies participate; recurring windows remain #216/#243).
- Tests: a group-only schedule/budget must (a) show up in the timekpra args of a push, (b) drive a force-close decision, (c) appear in the preview diff — each with a user-level rule overriding it per ADR 0004/0007 precedence.
- Confirm the unlink/unrestricted push (#253) and offline-queue behaviour are unaffected.
- One decision to record: `effective.ts` composes at the API layer while the push composes in the transport — after this change there should be a **single** composition point (or a shared helper) so the two can't drift again.

## Out of scope

- Exceptions in resolution (user *or* group) — `resolve.ts` doesn't consume exceptions at all yet; that's #142.
- Weekday-varying budgets (#141); UI (#343 and the group-editor follow-ups).

## References

- #124 (umbrella — its scope line "resolution helper … reused by both enforcement and the display surfaces" is the unfinished half), #182, #134, #98/#99, ADR 0004 / 0007 / 0008.

BenSeymourODB:
## Summary

- Group-targeted schedules/budgets (#134/#182) landed end-to-end **except enforcement**: the `gatherUserScheduleRules` / `gatherUserBudgets` resolution helpers were consumed only by the display endpoint `GET /api/users/:userId/effective`, so group rules were silently **display-only** — `/admin`'s effective view and the client's actual behaviour disagreed.
- Makes `policy/group-resolution.ts` the **single group-aware composition point** and routes every enforcement / preview / render path through it, so the enforced answer can't drift from the displayed one again.
- Group-only bedtimes/allowed-hours and group budgets now actually reach `timekpra`; group per-activity quotas now drive the force-close sweep; the save-and-push preview and the e2guardian/AppArmor renders include inherited always-on group denies — each with a user-level rule still overriding an inherited one per ADR 0004/0008 precedence.

## Plan

Linked plan: `.claude/plans/362-group-inherited-enforcement.md`
Roadmap: docs/roadmap.md → Phase 2 (policy model / group layer)

Delivered in two commits:
1. **Core enforcement** — extract `mergeScheduleRulesWithGroups` / `mergeBudgetsWithGroups` cores (own-rules passed explicitly); `gatherUser*` become thin wrappers. Route the policy-push executor + `enforcement/evaluate` through them. Widen `PolicyEnforcementContext` to the `ScheduleRule` / `BudgetInput` subsets the Linux runner already forwards to `resolvePolicyPush`.
2. **Preview + ansible renders** — preview `before` uses `gatherUser*`, `after` merges the *proposed* own rules with the persisted group layer; `e2guardian.resolveBannedSites` and `apparmor.listUserAlwaysOnDenies` read `gatherUser*`.

## License-boundary note

N/A — plain TypeScript over the policy model + Drizzle reads. Ansible and `timekpra` stay subprocesses; nothing linked/imported/vendored; no GPL binary added to the image.

## Test plan

- [x] Policy push: a group-only overall daily budget reaches `timekpra`; a user's own budget overrides it; a group always-on overall deny shapes the resolved push.
- [x] Force-close: a group-inherited per-activity budget fires; a user's own budget override suppresses it.
- [x] Preview: the diff includes an inherited group schedule window and an inherited group budget baseline on both sides (stale #182-era fidelity test rewritten to the corrected contract).
- [x] e2guardian / AppArmor: an always-on group deny appears in the banned-sites plan / the AppArmor denials.
- [x] `npm run format:check && npm run lint && npm run typecheck && npm test` — 1683 passing, coverage gate 80% held.

## Deferred (out of scope, per issue)

- Exceptions in resolution (user or group) — `resolve.ts` consumes no exceptions yet (#142).
- Weekday-varying group budgets (#141); group-editor UI (#343/#363).
- Recurring/date-scoped group denies in the static e2guardian/AppArmor renders (#216/#243) — only **always-on** group denies participate, matching the existing own-rule behaviour.

Closes #362

### DIFF
--- a/.claude/plans/362-group-inherited-enforcement.md
+++ b/.claude/plans/362-group-inherited-enforcement.md
@@ -1,0 +1,83 @@
+# Plan — #362: wire group-inherited schedules/budgets into enforcement
+
+**Issue:** #362 (Phase 2, `bug`). Group-targeted schedules/budgets landed
+end-to-end (#134/#182) *except enforcement*: the resolution helpers
+`gatherUserScheduleRules` / `gatherUserBudgets` (`policy/group-resolution.ts`)
+are consumed by exactly one caller — the display endpoint
+`GET /api/users/:userId/effective`. Every path that actually enforces or
+previews reads the user's **own** rows only, so group rules are silently
+display-only.
+
+## Root cause
+
+Two composition points exist and have drifted:
+
+- **Display** (`api/policy/effective.ts`) composes via the `gather*` helpers →
+  includes group rules.
+- **Enforcement / preview / ansible renders** read own-only rows
+  (`listUserSchedules` / `listUserBudgets` / inline `db.select()`), never the
+  group layer.
+
+## Fix — one shared composition point
+
+Make `policy/group-resolution.ts` the single entry point every surface reads,
+so display, push, force-close, preview, and the ansible renders can't drift
+again (the one decision the issue asks to record).
+
+Extract merge cores that take the *own* rules explicitly, so preview can merge
+a **proposed** own-rule set with the **persisted** group rules:
+
+- `mergeScheduleRulesWithGroups(db, userId, ownRules)` — own-first, then each
+  group's rules (ascending group id), re-sequenced to dense ordinals.
+- `mergeBudgetsWithGroups(db, userId, ownBudgets)` — own-first full-replace per
+  `(scope, window, target)` slot.
+- `gatherUserScheduleRules` / `gatherUserBudgets` become thin wrappers that pass
+  the user's persisted own rows to those cores (behaviour unchanged — existing
+  tests stay green).
+
+## Phase 1 — core enforcement paths (push + force-close)
+
+1. `policy/group-resolution.ts`: extract the two merge cores; keep `gatherUser*`
+   as wrappers.
+2. `transport/policy-push/platform-runner.ts`: widen
+   `PolicyEnforcementContext.schedules` from `readonly ScheduleRow[]` to
+   `readonly ScheduleRule[]`, and `budgets` from `readonly BudgetRow[]` to
+   `readonly BudgetInput[]`. The Linux runner only forwards these to
+   `resolvePolicyPush` (which already takes those subsets), so this is a safe
+   widening; `GatheredScheduleRule`/`GatheredBudget` satisfy the subsets.
+3. `transport/policy-push/executor.ts`: read `gatherUserScheduleRules` /
+   `gatherUserBudgets` instead of `listUserSchedules` / `listUserBudgets`.
+4. `enforcement/evaluate.ts`: replace the inline own-only schedule/budget
+   `db.select()`s with the `gather*` helpers (grants stay own-only — they are a
+   per-user additive layer).
+5. Tests: a group-only schedule/budget shows up in a push's `timekpra` args and
+   drives a force-close decision; a user-level rule overrides an inherited one.
+
+## Phase 2 — preview + ansible renders
+
+6. `api/policy/preview-routes.ts`: `before` = `gatherUser*` (mirrors the push);
+   `after` = `mergeScheduleRulesWithGroups(proposedOwn)` /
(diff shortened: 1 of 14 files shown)


BenSeymourODB:
🤖 First-pass self-review done (adversarial, against `CLAUDE.md` + the license boundaries).

**Verified clean:** no GPL boundary touched (pure TS over the policy model + Drizzle reads; `timekpra`/Ansible stay subprocess/config-file); the `gatherUser*` refactor preserves persisted-case behaviour exactly (own-first precedence, dense ordinal re-sequencing, per-slot full-replace budget override, group-id tiebreak, within-source summing); the `PolicyEnforcementContext` widening (`ScheduleRow→ScheduleRule`, `BudgetRow→BudgetInput`) is safe — the Linux runner only forwards both into `resolvePolicyPush`; grants stay own-only; no `any`/`as`/`@ts-ignore`.

**One minor finding, now fixed (`260baf7`):** `mergeScheduleRulesWithGroups` re-sequenced ordinals by array index, so the save-and-push preview — which passes proposed own rules in request-body order — could resolve own-rule precedence by array position instead of `ordinal`, diverging from the push (which persists the ordinal and reloads sorted). Fixed by sorting own rules with `byOrdinal` before the merge re-sequences them; added a unit test. The persisted path was already sorted, so it's unchanged.

The stale pre-#362 preview test that asserted "group rules are not pushed" was rewritten to the corrected contract (group windows/budgets now appear on both diff sides) — a correction of a test that pinned the bug, not a weakening.

Quality gate green locally: format, lint, `tsc --noEmit`, 1684 unit tests, coverage ≥80%.

---

Issue #327, opened 2026-06-27
Boot wiring: compose telemetry pull → UsageSample rollup → enforcement sweep in main.ts


BenSeymourODB:
## What

Wire the live periodic loop into `server/src/main.ts` so per-activity enforcement actually runs in a deployed dashboard:

1. Compose a real telemetry consumer that **fetches AW window/afk events, normalises them (#88 `normaliseWindowEvents`), and inserts `UsageSample` rows (`insertUsageSamples`)** into the pull's `consume()` seam — this is #88's explicitly-deferred "wire the normaliser into #162's scheduled pull" step.
2. Start `scheduleTelemetryPull` at boot (after `listen`, like the Ansible/AdGuard bootstraps), and **drive the enforcement sweep right after each rollup pass** by calling `EnforcementSweepHandle.tick()` (the sweep landed in #292 / PR for the sweep) — so enforcement reads fresh usage rather than racing the rollup on a separate timer.
3. Construct the single long-lived `ForceCloseTrigger` once at boot via `createForceCloseDeps({ db, eventHub, ssh, credentials, sink, logger })` and hand it to `startEnforcementSweep`.
4. Add env config for the cadence + cool-down (`PCT_*`), and stop both handles on `app.close()`.

## Why this is separate

The repo ships each periodic job as a seam-injected, unit-tested `start*` unit and defers the `main.ts` boot wiring — the offline-queue drainer (#84), periodic re-apply, and the telemetry pull (#162) are all built but unstarted. The enforcement sweep (#292) follows the same pattern. This issue is the one place that turns the whole telemetry → normalise → insert → enforce chain on, so it is best done as a single focused boot-wiring change once #88's consumer composition lands here.

## Blocked by / depends on

- #88's deferred "wire the normaliser into the scheduled pull `consume()` seam" (the rollup that writes `UsageSample`s).
- The enforcement sweep unit (#292 PR) providing `startEnforcementSweep` / `EnforcementSweepHandle`.

## Acceptance

- A real telemetry consumer writes `UsageSample`s on each scheduled pull.
- `tick()` of the enforcement sweep runs after each rollup; an exhausted daily per-activity budget produces an `enforce.force_close` (or SSH `pkill` fallback) end-to-end.
- Both schedulers start after `listen` and stop on shutdown; cadence/cool-down are env-configurable.
- Reuses the live SSH round-trip confidence gate (#157) before trusting periodic enforcement on a real client.

## References

[...]

PR #397, opened 2026-07-27
feat(telemetry): durable per-client pull cursor (#382)

### LINKED ISSUE #382: Durable per-client telemetry pull cursor (survive restart)
## What

Promote the telemetry pull cursor from the **in-memory** map introduced in #327
to a **durable** per-client cursor that survives a dashboard restart.

## Background

`insertUsageSamples` is a plain append with no uniqueness constraint, and cross-pull
de-duplication was always the pull layer's responsibility (#162 deferred the cursor).
When #327 turned on real `UsageSample` inserts it added a minimal **in-memory**
`Map<clientId, Date>` cursor: each pass queries `[cursor ?? (passEnd − initialLookback), passEnd]`
and advances the cursor to `passEnd` only after a successful insert, so overlapping
passes within one process lifetime don't double-count
(`server/src/enforcement/telemetry-consumer.ts`).

The gap: the cursor is not persisted, so after a restart each client's first pass
re-pulls the `PCT_ENFORCEMENT_INITIAL_LOOKBACK_SECONDS` window (default 900 s). That
window can overlap already-persisted samples → a bounded double-count, or (if the
downtime exceeds the lookback) a gap. Missing telemetry credits no consumption (#88),
so a gap is non-punitive, but the restart overlap is a real correctness wrinkle.

## Scope

- Persist the last successfully-pulled window `end` per client (a column on `clients`,
  e.g. `last_telemetry_pull_at`, or a small `telemetry_cursors` table). Timestamp-prefixed
  Drizzle migration (`npm run db:generate`, `CLAUDE.md` #133).
- Seed the in-memory cursor from the persisted value on boot; advance both together on a
  successful insert. Keep the `initialLookback` seed only for a client with no persisted
  cursor yet.
- Alternatively/additionally: make the append idempotent (a natural key or an
  overlap-aware upsert) so a re-pull can't double-count regardless of cursor fidelity —
  decide which layer owns the guarantee.

## References

- Follow-up to #327 (boot wiring; introduced the in-memory cursor).
- `server/src/enforcement/telemetry-consumer.ts`, `server/src/policy/usage.ts`
  (`insertUsageSamples` — "the telemetry-pull layer owns the per-client pull cursor").
- `docs/architecture.md` → "Inbound (client → server) — telemetry pull".

BenSeymourODB:
## Summary

- Promote the Phase-5 telemetry pull cursor from the **in-memory** `Map` introduced in #327 to a **durable** per-client cursor persisted on `clients.last_telemetry_pull_at`, so a dashboard restart resumes each client's pull exactly where the last successful pull ended instead of re-pulling the whole `PCT_ENFORCEMENT_INITIAL_LOOKBACK_SECONDS` window (a bounded double-count against already-persisted samples).
- The pipeline seeds its in-memory cursor from the column on boot; the consumer persists each advance at the **same** success point it advances the in-memory cursor, so a mid-pull failure leaves both unmoved and the window re-pulls next pass. A client with no persisted cursor still falls back to `initialLookback`.

## Design note — column vs table

Persisted as a nullable timestamp column on `clients` (not a separate `telemetry_cursors` table): the cursor is strictly 1:1 with a client, dies with it (no orphan rows, no extra FK), and mirrors the existing `last_seen` / `versions_reported_at` nullable-timestamp columns. `NULL` = "no successful pull yet".

The issue's *"alternatively/additionally"* idempotent-insert option was **not taken**: adding a uniqueness constraint / overlap-aware upsert to `usage_samples` is a larger change with its own natural-key question and would alter the deliberately-plain append. Cursor persistence fixes the documented restart overlap with a single nullable column and no change to the write path.

## Plan

Linked plan: `.claude/plans/issue-382-durable-telemetry-cursor.md`
Roadmap: docs/roadmap.md → Phase 5 (ActivityWatch telemetry pull)

## License-boundary note

N/A — Drizzle (Apache-2.0) + better-sqlite3 (MIT) reads/writes only. No GPL source linked in-process, no GPL binary added to the image, no transport or packaging change.

## Test plan

- [x] `policy/telemetry-cursor.test.ts` — `saveTelemetryCursor` writes the column; `loadTelemetryCursors` returns only clients with a persisted value, keyed by id; overwrite/advance; per-client independence; empty DB → empty map.
- [x] `enforcement/telemetry-consumer.test.ts` — a successful pull (incl. a clean pull that matched nothing) persists the cursor to the `clients` row; a failed pull leaves it `NULL` (both cursors unmoved).
- [x] `enforcement/pipeline.test.ts` — a client with a persisted cursor seeds the in-memory cursor so the first pass's window `start` resumes at the persisted `end`, not `passEnd − initialLookback`.
- [x] Full gate green from `server/`: `format:check`, `lint`, `typecheck`, `npm test` (1876 tests, coverage 98.9% ≥ 80%).

Closes #382

---
_Generated by [Claude Code](https://claude.ai/code/session_01J7yoTLkNNgCU8jEvoJAL3v)_

### DIFF
--- a/.claude/plans/issue-382-durable-telemetry-cursor.md
+++ b/.claude/plans/issue-382-durable-telemetry-cursor.md
@@ -1,0 +1,87 @@
+# Plan — #382 Durable per-client telemetry pull cursor (survive restart)
+
+Roadmap: `docs/roadmap.md` → Phase 5 (ActivityWatch telemetry pull).
+Issue: #382 (follow-up to #327, which introduced the **in-memory** cursor).
+
+## What exists on `main` (don't rebuild)
+
+- `enforcement/telemetry-consumer.ts` — `createUsageTelemetryConsumer` holds an
+  **in-memory** `Map<number, Date>` cursor (last successfully-pulled window
+  `end`, keyed by `clientId`). Each pass queries
+  `[cursor.get(id) ?? (passEnd − initialLookback), passEnd)` and, only after a
+  successful `insertUsageSamples`, advances `cursor.set(id, end)`. A mid-pull
+  throw leaves the cursor unmoved so the window re-pulls next pass.
+- `enforcement/pipeline.ts` — owns the cursor: `const cursor = new Map()` at
+  construction, seeds the consumer + the sweep, pins `currentPassEnd` per pass.
+- `policy/usage.ts` — `insertUsageSamples` is a plain append (no uniqueness
+  constraint); cross-pull de-dup is explicitly the pull layer's job.
+- `policy/schema.ts` — `clients` table (`last_seen`, `versions_reported_at`,
+  … all nullable `integer(mode:"timestamp")` columns to mirror).
+
+## The gap this closes
+
+The cursor is process-local, so after a restart a client's first pass re-pulls
+the whole `PCT_ENFORCEMENT_INITIAL_LOOKBACK_SECONDS` window (default 900 s),
+which overlaps already-persisted samples → a bounded restart double-count.
+Persisting the cursor makes the first post-restart window start exactly where
+the last successful pull ended: no overlap, no gap.
+
+## Design decision — column vs table
+
+Persist on **`clients`** as a nullable `last_telemetry_pull_at` timestamp
+column, not a separate `telemetry_cursors` table. The cursor is strictly 1:1
+with a client, dies with the client (no orphan rows, no extra FK/cascade), and
+mirrors the existing `last_seen` / `versions_reported_at` nullable-timestamp
+columns. `NULL` is the "no cursor yet" state → fall back to `initialLookback`.
+
+**Idempotent-insert alternative (issue's "alternatively/additionally"): not
+taken.** Adding a uniqueness constraint / overlap-aware upsert to
+`usage_samples` is a larger change with its own semantics question (what is the
+natural key of a clipped interval?) and would alter the deliberately-plain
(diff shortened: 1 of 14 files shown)


BenSeymourODB:
🤖 First-pass review done (self-review before marking ready). It surfaced one real correctness gap, now fixed in `fe54753`:

- **Atomicity (fixed).** The durable `saveTelemetryCursor` and the `UsageSample` insert were two independently auto-committed writes. A crash in the window between them (SIGKILL / OOM / power loss) could leave the cursor *behind* already-persisted samples, so the next boot re-pulls that window and appends its rows a second time (`usage_samples` has no uniqueness constraint, by design) — the punitive double-count this feature removes. Both writes are now in one `db.transaction()` (the `policy/enrolment.ts` idiom), and the in-memory cursor advances only after it commits, so it can never outrun the durable one. Added `PolicyTx` to `policy/db.ts` and widened the two write helpers to `PolicyDb | PolicyTx` (a safe param widening).
- **Test hardening (fixed).** Added a case asserting a *pre-existing* durable cursor is left unchanged when a pass fails (`T1 → fail → still T1`), not just the `null → null` case.
- **Skipped (no change):** a suggestion to inject `loadTelemetryCursors` as a pipeline seam. It's already exercised through the real function against a seeded DB and over-injecting would just add noise — the reviewer flagged it as not worth changing, and I agree.

Full gate green after the fix: `format:check`, `lint`, `typecheck`, `npm test` (1877 tests, coverage 98.9% ≥ 80%).

---

Issue #400, opened 2026-08-07
Admin Clients view: surface per-client supported capabilities (grey out unsupported controls)


BenSeymourODB:
## What

Follow-up to **#288**, which landed the backend capability gate: the
`EventHub` now records each connection's negotiated `capabilities` and
withholds a gated frame (`enforce.force_close` → `per_app_close`,
`enforce.session_lock` / `lockout.cleared` → `session_budget`) from any
connection that didn't advertise the matching capability (ADR 0007 §4).

#288 deliberately deferred bullet 3 of the issue — the **admin surface** — so
the correctness core could land as a focused, fully-unit-tested slice without
adding a schema column no UI reads yet.

## Scope

- **Persist** the negotiated `capabilities` on the `Client` row at handshake
  `accept` (alongside the existing `recordClientAgentVersion` write in
  `events/stream.ts`), so they survive beyond the live connection. Drizzle
  migration via `npm run db:generate` (timestamp-prefixed, #133).
- **Expose** them in the Clients / health DTO (`api/clients/health-*.ts`).
- **Surface** per-client supported capabilities in the admin Clients view and
  **grey out unsupported controls** per client — the Windows-client modularity
  seam described in `docs/windows-client-support.md` → "Modularity tweaks"
  (a Linux client and a future Windows client speak the same `eventProtocol`
  and differ only in advertised `capabilities`).

## References

- #288 (the live per-connection gate this builds on),
  `docs/adr/0007-event-stream-version-compatibility.md` §4,
  `docs/windows-client-support.md`.
- The capability vocabulary lives in `server/src/events/capabilities.ts`
  (`CLIENT_CAPABILITIES`).

Phase 8b.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Picked after a full phase-ordered triage of the open backlog. Earlier-phase items were skipped as ineligible for an end-to-end single-session run:

- **Phase 3** (#392–395): chained behind open PR #411.
- **Phase 4**: #157 needs a live `timekpra` container to validate the CLI round-trip (can't be exercised here); #281's remaining bullets are gated (future-dated preview on the resolver, Ansible diff on Phase 6) and it was triaged hours ago; #318 is ADR-gated (#314 open); #185 is an epic, #261 a manual run.
- **Phase 5** #369: the issue directs co-designing the per-user health axis *with* #103, not solo.
[...]

Issue #165, opened 2026-06-18
Version handshake + N-1 compatibility window on the event stream / /api


BenSeymourODB:
## What

Define and enforce a **version-compatibility contract** between the dashboard and the fleet, built into the `pct-client-bridge` ↔ `/api/events/stream` handshake (and aligned with the existing `/api/meta` `apiVersion`). This is the lynchpin that makes "upgrade the server first, roll the clients afterwards" safe (#163): a freshly-upgraded server must keep talking to clients that haven't updated yet.

## Why this phase (slot in sooner)

The WebSocket bridge and the event stream are built in Phase 8b. Negotiating a protocol version *in the handshake from day one* is trivial; bolting it on after clients are deployed without it is not (an un-versioned client can't be told how to behave). `docs/client-notifications.md` already notes the WS wire format is fixed during implementation — fix the version field at the same time.

## Scope

- On connect, the bridge sends its agent version + a supported event-protocol version; the server records it (feeds #ENROL-VERSIONS) and selects a frame dialect it and the client both understand.
- Publish a **support policy**: the server supports clients within a documented window (e.g. current and previous minor, "N-1"). Outside the window → the server refuses the stream with a typed error and flags the client as `update_required` in the admin inventory rather than sending frames the client can't parse.
- Make the event-frame schema explicitly versioned (additive changes don't bump; breaking changes do) and document it in the API reference alongside the `update.*` events (#UPDATE-EVENTS).
- Mirror the same `apiVersion` discipline on `/api/*` so external integrators (#163 design rules) and clients read one consistent contract.

## Out of scope

- The actual update mechanism (Ansible playbook, orchestration) — Phase 14 (#163).

## Dependencies

- Built with the event stream + bridge (Phase 8b, see `docs/client-notifications.md`).
- Underpins every Phase 14 rollout (#163): the compatibility window defines how far the fleet can lag the server.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

ADR [0007](https://github.com/BenSeymourODB/linux-parental-controls-toolkit/blob/main/docs/adr/0007-event-stream-version-compatibility.md) has settled the contract; this run implements the **server side** against §7's target. The bridge client (#101) doesn't exist yet, but the handshake is exercisable end-to-end by simulating a client `hello` in tests, and adding a mandatory `hello` is safe because no bridge is deployed.

Scope for this PR:
[...]

PR #401, opened 2026-08-07
feat(events): capability-based frame gating on the event stream (#288)

### LINKED ISSUE #288: Event stream: capability-based frame gating (#165 follow-up, ADR 0007 §4)
## What

Follow-up to **#165** (ADR 0007), which landed the event-stream version handshake: the `negotiate` N-1 window, the `hello`/`accept`/`refuse` handshake on `GET /api/events/stream`, `/api/meta`'s `eventProtocol`, and the persisted `update_required` flag.

#165 deliberately deferred the **capability-based frame gating** of ADR 0007 §4:

> Because additive changes do not bump `eventProtocol`, the mechanism for "the server has a new event type that only newer clients understand" is the `capabilities` set in the handshake, not the version window. The server only sends a frame type a client advertised support for.

Today the handshake **accepts and could record** a client's `capabilities`, but nothing gates outbound frames on them — because there are no per-capability frame producers yet. This issue adds the gating once the producers exist.

## Scope (once the producers land)

- Persist / make available the negotiated `capabilities` per connection (and/or per client).
- In the hub's publish path, withhold an `enforce.*` / `lockout.*` frame from a connection that did not advertise the matching capability (e.g. `session_budget`, `per_app_close`), per ADR 0007 §4 — an older or differently-capable client simply never receives a frame it can't honour, and is **not** refused for it.
- Surface per-client supported capabilities in the admin Clients view so unsupported controls can be greyed out (the Windows-client modularity seam, `docs/windows-client-support.md`).

## Dependencies

- The `enforce.*` frame producers (#99) and the lockout flow (#108) — there is nothing to gate until they emit frames.
- Builds on #165 (handshake + `capabilities` in the `hello`).

## References

- `docs/adr/0007-event-stream-version-compatibility.md` §4.
- `docs/windows-client-support.md` → "Modularity tweaks".

BenSeymourODB:
## Summary

- The `/api/events/stream` handshake already negotiates each client's `capabilities` set (#165), but the `EventHub` discarded them and fanned every frame out to every connection. Per **ADR 0007 §4**, the server should only send a frame type a client advertised support for.
- Adds `events/capabilities.ts` — the capability vocabulary (`per_app_close`, `session_budget`) and a **pure, exhaustive** `capabilityForEvent` map. `enforce.force_close` → `per_app_close`; `enforce.session_lock` / `lockout.cleared` → `session_budget`; `grant.applied` / `policy.changed` stay ungated. The exhaustive switch makes a new frame type a compile error until its gate is declared.
- `EventHub` records each connection's advertised capabilities and **withholds a gated frame from any connection lacking the matching capability** (both `publishToClient` and `broadcast`); an older / differently-capable client simply never receives a frame it can't honour, and is **not** disconnected for it. `stream.ts` threads the accepted `hello.capabilities` into `hub.register`.

## Plan

Linked plan: `.claude/plans/288-event-capability-gating.md`
Roadmap: docs/roadmap.md → Phase 8b. Follow-up to #165 (ADR 0007).

## Deferred (tracked)

- **Surface per-client supported capabilities in the admin Clients view** (persist on the `Client` row + DTO + grey-out UI — the Windows-client modularity seam) → filed as **#400**. Adding an unused `Client.capabilities` column here would be dead code until the UI consumes it; the live per-connection gate in this PR is the load-bearing correctness deliverable.

## License-boundary note

N/A for GPL boundaries — pure TypeScript event-hub logic + Fastify/`@fastify/websocket` (MIT). No GPL linkage, no transport/packaging change, no image change. No new dependencies.

## Test plan

- [x] `tests/events/capabilities.test.ts` — the `capabilityForEvent` mapping + the `CLIENT_CAPABILITIES` wire-string values.
- [x] `tests/events/hub.test.ts` — new `capability gating` suite: withheld without the cap, delivered with it, ungated frames reach everyone, per-connection filtering within one client, `delivered` counts only recipients, broadcast gating, one capability doesn't authorise another, re-register refreshes caps. (Existing fan-out/stamping tests that use the now-gated `lockout.cleared` advertise `session_budget` — a faithful adaptation, not a weakening.)
- [x] `tests/events/stream.test.ts` — end-to-end: a client advertising only `session_budget` receives `enforce.session_lock` but not `enforce.force_close`, verifying the handshake→hub capability threading through the real route.
- [x] Full gate green from `server/`: `format:check`, `lint`, `typecheck` (`tsc --noEmit`), `vitest run` with the 80% coverage gate.

Closes #288

---
_Generated by [Claude Code](https://claude.ai/code/session_0135iXaCUhombHP3eABPmuBZ)_

### DIFF
--- a/.claude/plans/288-event-capability-gating.md
+++ b/.claude/plans/288-event-capability-gating.md
@@ -1,0 +1,88 @@
+# Plan — #288 Event stream: capability-based frame gating (ADR 0007 §4)
+
+Roadmap: Phase 8b. Follow-up to #165 (event-stream version handshake).
+ADR: `docs/adr/0007-event-stream-version-compatibility.md` §4.
+
+## Problem
+
+The `/api/events/stream` handshake (#165) already negotiates a client's
+`capabilities` set (`events/protocol.ts` → `helloFrameSchema.capabilities`),
+but nothing gates outbound frames on it. On `accept`, `stream.ts` calls
+`hub.register(clientId, socket)` and **discards** the advertised capabilities;
+the `EventHub` fan-out (`publishToClient` / `broadcast`) sends every frame to
+every open connection regardless of what the client can honour.
+
+ADR 0007 §4: *"The server only sends a frame type a client advertised support
+for; an older client simply never receives a frame it couldn't handle, and is
+**not** refused for it."* The `enforce.*` producers already exist (#99), so the
+gate is now implementable and testable end-to-end.
+
+## Scope (this PR — the backend correctness core)
+
+1. **Capability vocabulary + frame→capability map** — new
+   `server/src/events/capabilities.ts`:
+   - `CLIENT_CAPABILITIES` constants: `per_app_close`, `session_budget`
+     (the ADR §4 examples with frame producers today; `applocker_deny` /
+     `dns_filter` are forward-looking and add trivially when their frames land).
+   - `capabilityForEvent(event): ClientCapability | null` — a **pure,
+     exhaustive** switch over `ServerEventType`. A new frame type fails to
+     compile until its gate is declared (`assertNever`). Mapping:
+     | event | required capability |
+     | --- | --- |
+     | `enforce.force_close` | `per_app_close` |
+     | `enforce.session_lock` | `session_budget` |
+     | `lockout.cleared` | `session_budget` |
+     | `grant.applied` | — (baseline, ungated) |
+     | `policy.changed` | — (baseline, ungated) |
+
+     `lockout.cleared` gates on `session_budget`: it only clears an
+     overall-budget lockout, which can only happen on a client doing
+     session-budget enforcement (matches this issue's `enforce.*`/`lockout.*`
+     wording). `grant.applied` / `policy.changed` are informational nudges every
+     client re-renders, so they stay ungated.
+
+2. **Hub gating** — `server/src/events/hub.ts`:
+   - Store each connection's advertised capabilities alongside the socket
+     (`Map<EventSocket, ReadonlySet<string>>` per client).
+   - `register(clientId, socket, capabilities = [])` records them; default `[]`
+     = advertises nothing = receives only ungated frames (the correct default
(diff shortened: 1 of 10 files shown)


BenSeymourODB:
First-pass self-review done (deep read of the diff against ADR 0007 §4, `CLAUDE.md` license boundaries, and the touched tests). Outcome:

**Addressed — cross-module invariant (commit `bd83797`).** The gate makes `EventHub.publishToClient` return `0` in a second case: a client that is *connected but did not advertise `per_app_close`*, so the hub withholds the frame. `enforcement/force-close.ts` reads that delivery count as its dispatch signal and, on `0`, falls back to the SSH `pkill`. That fallback is the **correct** enforcement path for such a client (its agent can't act on a frame it can't parse, and `pkill` is server-side / capability-independent), and it's audited as the SSH path it actually took — so **no behaviour change was needed**. What was stale was the module's documented invariant ("`0` means the bridge is gone"); refined it and its collaborator docs so the contract `force-close.ts` depends on is accurate now that the gate exists. Latent on the Linux MVP (always advertises `per_app_close`), real for the capability-differentiated future this gate exists to serve.

**Ratified, no change:**
- Gating `lockout.cleared` on `session_budget` goes one frame beyond the ADR's literal "`enforce.*`" wording, but on a clean invariant — a client that could never receive `enforce.session_lock` can never be in a locked state, so it never needs the clear. Documented in the plan; matches this issue's `enforce.*`/`lockout.*` framing.
- `capabilityForEvent`'s `default → assertExhaustive(never)` runtime throw is an unreachable defensive guard; the real protection is the compile-time exhaustiveness (a new `ServerEvent` variant won't compile until its gate is declared). Global coverage gate passes.

License boundaries clean (pure TS, no GPL linkage, no image change); no `any`/`as`/`@ts-ignore`/non-null-`!` introduced. Local gate green (`format:check`, `lint`, `typecheck`, `vitest` 1891 passing + 80% coverage); CI green bar the unit-tests job still finishing.

---

Issue #137, opened 2026-06-17
Retention: scheduled automatic purge job (croner) + audit of each run


BenSeymourODB:
## What

The croner-scheduled job that **automatically** enforces the configured retention windows (#sub-1), per the epic #135.

## Scope

- A periodic (e.g. daily) croner job that, per category, deletes records older than the configured window using the shared `isExpired` helper.
- Batched/bounded deletes so a large first run doesn't lock the store; safe to interrupt and resume.
- **Audit every run**: record what category, how many rows, and the cutoff in the audit log (#85) — purges must be observable.
- Dry-run / preview mode (count what *would* be purged) for operator confidence; surface last-run summary in the admin retention page.
- Manual "run now" trigger from the admin UI.

## Notes

- The job only *executes* policy; the per-entity deletion rules (what counts as purgeable, FK/ledger constraints) live in #sub-3.
- Must respect the Grant-ledger constraint from the epic (don't orphan a revocation from its grant).

Part of #135.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Scoping this run to the full **scheduled automatic purge job (#137)** on top of the merged config (#136) and per-entity purge coverage (#138):

- A retention **purge service** that loads the effective `RetentionPolicy` (env default + persisted overrides), drives `purgeExpiredRecords`, and **audits each run** (per-category cutoff + rows deleted) to the audit log (#85).
- A **dry-run / preview** mode (counts what *would* be purged without deleting) — adding a `countExpiredRecords` counterpart to `purge.ts`.
- A **croner scheduler** (daily default, env-overridable) with overlap protection, wired at boot alongside the existing schedulers (constructed in `buildApp`, started in `main.ts`, stopped on `onClose`) — mirroring the offline-queue / reapply / enforcement-pipeline pattern.
- **API**: manual "run now" + dry-run preview + last-run summary under `/api/retention`.
- **Admin retention page**: surface the last-run summary and a "run now" control.

Respects the epic's grant-ledger constraint (a revocation is a column on the grant row, so purging can't orphan it) and the "keep forever ⇒ never purged" escape hatch, both already honoured by the #138 routines. License boundary untouched — pure TypeScript + Drizzle over the policy store.

Plan + tests to follow on a draft PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #136, opened 2026-06-17
Retention: policy model + configuration (default 1 year, per-category overrides, admin surface)


BenSeymourODB:
## What

Define the data model and configuration for retention windows, with a **1-year default**, per the epic #135.

## Scope

- A retention configuration: a default window (365 days) plus optional per-category overrides (e.g. `usage_samples`, `schedule_history`, `budget_history`, `exceptions`, `audit_log`, `grant_ledger`), each able to be set to a custom duration or "keep forever".
- Persist config in the policy store (or env-overridable defaults documented in `.env.example` / `docs/server-deployment.md`); zod-validated.
- `/api/*` read/write for the config behind the admin guard; admin UI surface to view/adjust windows with the default clearly shown.
- A pure helper `isExpired(record, category, now)` reused by the purge job (#sub-2) so the rule lives in one place.

## Notes

- Default and escape hatch behaviour follow the epic's principles (conservative; "keep forever" allowed).
- Timestamps compared in UTC ([ADR 0001](../blob/main/docs/adr/0001-budget-timezone.md)).

Part of #135.

BenSeymourODB:
🤖 implement-issue claiming this for the next session.

Scoping this run to the **backend slice** of #136: the retention-config data model, persistence, validation, the read/write `/api/*` endpoints behind the admin guard, and the pure `isExpired(record, category, now)` helper (UTC, per ADR 0001) that the purge job (#137/#138) will reuse.

**Deferred (will be tracked by a follow-up issue, linked from the PR):** the dedicated **admin UI surface** for viewing/adjusting windows. The admin shell + editors are themselves in flight (#53), so building the UI now would race that work; the JSON API + DTOs land here so the UI is a thin follow-up. The API will be exercised by tests directly.

Approach: env-overridable defaults (documented in `docs/server-deployment.md`) as the baseline, with DB-persisted per-category overrides; zod-validated DTOs shared with the frontend; one timestamp-prefixed Drizzle migration generated against current `main`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #135, opened 2026-06-17
Epic: data retention policies + automatic enforcement (default: purge records older than 1 year)


BenSeymourODB:
## What

Introduce **data retention policies** and **automatic enforcement** of them, so the dashboard does not accumulate stale records indefinitely. Default: **do not retain records older than 1 year.** Primary target is schedule/budget-related history, with the high-volume neighbours (usage telemetry, audit) handled under the same machinery.

This is an **epic**; the implementation chunks are tracked as sub-issues below.

## Motivation

- A long-running household instance accrues unbounded history (usage samples from Phase 5 especially, plus expired exceptions/grants, superseded policy/budget/schedule entries, audit rows). Without retention the SQLite store grows forever and old behavioural data lingers longer than anyone needs.
- A sane, **configurable** default (1 year) keeps the store bounded and is a reasonable privacy posture for minors' data.

## Principles / constraints

- **Configurable, with a 1-year default.** Allow per-category overrides (e.g. keep audit longer than telemetry) and an explicit "keep forever" escape hatch.
- **Automatic.** A scheduled job enforces it without manual action; every purge run is itself recorded in the audit log.
- **Respect the immutable Grant ledger.** Grants/revocations are an append-only ledger (`docs/architecture.md` → External integrations). Purging *aged-out* ledger rows is a retention action, not an edit — but it must be deliberate and configurable, and a revocation row must never be purged while its target grant is retained. Default retention for the ledger should be conservative.
- **Conservative deletion.** Never purge records still referenced by an active window/budget computation; purge by age of the record's relevant timestamp (UTC), interpreted consistently with the budget-window rules ([ADR 0001](../blob/main/docs/adr/0001-budget-timezone.md)).

## Sub-issues

- [ ] Retention policy model + configuration (default 1y, per-category overrides, admin surface)
- [ ] Scheduled automatic purge job (croner) + audit of each purge run
- [ ] Retention coverage for schedule / budget / exception history (+ usage samples)

## References

- `docs/architecture.md` → "Policy model" and "External integrations" (Grant ledger immutability).
- Interacts with Phase 5 telemetry (`UsageSample` volume) and the Phase 4 audit log (#85).
- Suggested label/milestone: Phase 11 (hardening & polish); revisit timing once Phase 5 telemetry lands and the store starts growing.

PR #409, opened 2026-08-07
feat(policy): per-entity retention purge coverage (#138)

### LINKED ISSUE #138: Retention: per-entity purge coverage for schedule / budget / exception history (+ usage samples)
## What

Implement the per-entity deletion logic — *what* is purgeable and *how*, without breaking referential integrity or active computations — for the categories the epic #135 targets. This is the substantive chunk; the config (#sub-1) and the scheduler (#sub-2) drive it.

## Coverage & rules to define

- **Schedule / Exception history:** purge expired `Exception` rows past the window; purge superseded/soft-deleted `Schedule` rows (define what "history" means — depends on whether policy edits are versioned). Never purge a rule that is still active/current.
- **Budget history:** purge superseded/soft-deleted `Budget` rows past the window; never the currently-effective baseline.
- **Usage samples:** purge `UsageSample` rows older than the window (likely the highest-volume category from Phase 5). Ensure no active burndown/rollup still needs them.
- **Audit log & Grant ledger (careful):** apply only the configured (typically longer/forever) windows; honour the immutability constraint — a revocation row is not purged while its grant is retained.

## Scope

- One deletion routine per category, each respecting FK constraints and the "still-referenced ⇒ keep" rule.
- Idempotent and safe under the batched job (#sub-2).
- Tests per category: a record past the window is purged, an active/referenced one is kept, and ledger pairing is preserved.

## Notes

- If "schedule/budget history" turns out to require **versioning** of policy edits that doesn't exist yet, raise that as a prerequisite decision on #135 before building — it may itself warrant a small follow-up issue.

Part of #135.

BenSeymourODB:
## Summary

- Adds the **per-entity deletion routines** the retention epic (#135) calls for — one bounded, idempotent purge per retention category in `server/src/policy/purge.ts`, driven by the shared `RetentionPolicy` rule (#136) so nothing re-implements the age comparison.
- Each category keys its "age" on the **end** of the record's relevant window, so a purge only ever removes data wholly in the past — an active or future-dated record can never be selected:
  - `usage_samples` → `ended_at` older than the window
  - `date_overrides` → `exceptions`/`group_exceptions` past `expires_at`, and date-scoped `schedules`/`group_schedules` past `effective_to`; **open-ended recurrence rules (null `effective_to`) are never purged** (ADR 0005 §4)
  - `audit_log` → `at` older than the window
  - `grant_ledger` → `expires_at` older than the window, so an active grant is never purged; a **revocation is a column on the grant row, not a separate ledger row**, so purging can never orphan a revocation from its grant
- `purgeExpiredRecords` runs every category and returns a per-category `{ cutoff, deleted }` summary — the shape #137's scheduled job will audit per run.

## Plan

Linked plan: `.claude/plans/issue-138-retention-per-entity-purge.md`
Roadmap: `docs/roadmap.md` → Phase 11 (epic #135). Follow-up to #136 (retention model/config/API); unblocks #137 (scheduled purge job).

### Notable design decisions

- **Cutoff, not row-by-row.** `isExpired` is monotonic in the record timestamp, so each category resolves to a single cutoff instant and the purge is a set-based `DELETE WHERE <ts> < cutoff` (strict `<`, matching the rule). New pure helper `retentionCutoff` + `RetentionPolicy.cutoffFor` keep that derivation beside the rule it mirrors, with a boundary test asserting `isExpired(ts) ⇔ ts < cutoff` so the two never drift.
- **Bounded / interruptible batching.** better-sqlite3's bundled SQLite is not built with `SQLITE_ENABLE_UPDATE_DELETE_LIMIT`, so `DELETE … LIMIT` is unavailable. Each pass deletes the rows whose PK is in a `LIMIT batchSize` sub-select, looping until a short/empty batch. Each pass is its own implicit transaction, so a large first run never holds a long write lock and an interrupted run resumes cleanly (the cutoff predicate is stable).
- **"schedule/budget history" is not a category.** This issue's body named `schedule_history` / `budget_history`, but #136 already fixed the taxonomy: schedules/budgets are live recurrence rules, not dated history (ADR 0005 §4). No such tables exist, so nothing to purge there.

## License-boundary note

N/A — pure TypeScript + Drizzle (better-sqlite3 MIT, drizzle Apache-2.0) reads and deletes over the policy store. No GPL imports, no GPL binaries added to the image, no subprocess/REST boundary crossed or collapsed. No new dependency.

## Test plan

- [x] `tests/policy/purge.test.ts` (20) — per category: a row past the window is purged, an active/future-dated/still-referenced one is kept; open-ended recurrence rules never touched; grant active-grant + immutability invariants; date_overrides sums across all four tables; batching is bounded, exact, and idempotent (a second run is a no-op); `keepForever` ⇒ `cutoff: null, deleted: 0`; `purgeExpiredRecords` returns one result per category in declaration order.
- [x] `tests/policy/retention.test.ts` — added `retentionCutoff` / `cutoffFor` cases incl. the boundary equivalence with `isExpired`.
- [x] Full gate green from `server/`: `format:check`, `lint`, `typecheck`, `test` — **1899 unit tests pass**, coverage gate (80%) held; `purge.ts` / `retention.ts` at 100%.

## Deferred (tracked)

- The croner-scheduled job that *drives* these routines, per-run audit-log entries, dry-run/preview counts, and the manual "run now" trigger — **#137** (this PR gives it the deletion primitives + per-category summary it needs).

Closes #138

---
_Generated by [Claude Code](https://claude.ai/code/session_01F1Nz1ejvHUYTdRMvYdrf3n)_

### DIFF
--- a/.claude/plans/issue-138-retention-per-entity-purge.md
+++ b/.claude/plans/issue-138-retention-per-entity-purge.md
@@ -1,0 +1,148 @@
+# Plan — #138 Retention: per-entity purge coverage
+
+Roadmap: `docs/roadmap.md` → Phase 11 (epic #135). Follow-up to **#136**
+(retention model: `resolveRetention` / `isExpired` / `RetentionPolicy`, the
+`retention_overrides` config, and `/api/retention`). Unblocks **#137** (the
+croner-scheduled job that *drives* these routines).
+
+## What this delivers
+
+One bounded, idempotent **deletion routine per retention category**, each
+driven by the shared `RetentionPolicy` rule so nothing re-implements the age
+comparison. The categories are already fixed by #136 (`policy/enums.ts`,
+grounded in ADR 0005 §4 — recurrence rules are *not* dated data):
+
+| Category         | Table(s)                              | Age keyed on            | "still active ⇒ keep" guard |
+| ---------------- | ------------------------------------- | ----------------------- | --------------------------- |
+| `usage_samples`  | `usage_samples`                       | `ended_at`              | interval must be wholly past (`ended_at < cutoff`) |
+| `date_overrides` | `exceptions`, `group_exceptions`      | `expires_at`            | `expires_at < cutoff ≤ now` ⇒ already expired |
+|                  | `schedules`, `group_schedules`        | `effective_to`          | **only** rows with a non-null `effective_to`; open-ended recurrence rules are never purged |
+| `audit_log`      | `audit_log`                           | `at`                    | n/a (no dependents; FKs are `set null`) |
+| `grant_ledger`   | `grants`                              | `expires_at`            | `expires_at < cutoff ≤ now` ⇒ already expired (inactive) |
+
+### Why these age keys are safe
+
+- **`usage_samples` → `ended_at`.** A sample covers `[started_at, ended_at)`.
+  Keying on the *later* bound means a sample is purged only once its whole
+  interval is older than the window, so no in-window burndown/rollup can still
+  need it.
+- **`date_overrides` → the end of the override's active window.** An exception
+  is active during `[effective_from ?? created_at, expires_at)`; a date-scoped
+  schedule during `[effective_from, effective_to)`. Keying on the *end*
+  (`expires_at` / `effective_to`) means a rule is purged only when it is
+  *wholly in the past* — an active or future-dated override
+  (`expires_at`/`effective_to ≥ now`) can never satisfy `end < cutoff ≤ now`.
+  Schedules with a **null** `effective_to` are open-ended recurring rules (ADR
+  0005 §4) and are **never** matched (the predicate requires a non-null
+  `effective_to`). Group variants (`group_exceptions` / `group_schedules`)
+  carry the same columns and are covered identically — the resolver merges
+  user + group rows, so retention must too.
+- **`grant_ledger` → `expires_at`.** In this schema revocation is a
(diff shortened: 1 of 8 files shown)


BenSeymourODB:
First-pass review done (self-review via subagent). It confirmed the purge predicates, the `isExpired`↔cutoff boundary equivalence, the batched-DELETE termination/idempotency, and the grant-ledger immutability + open-ended-recurrence claims, with the full gate green. Two must-fix items were addressed in `fc2d739`:

1. Corrected the `grant_ledger` docstring in `enums.ts` — it now documents the shipped `expires_at` age key (was stale on `granted_at`), with the active-grant / revocation-column rationale.
2. Guarded `purgeInBatches` against a non-positive `batchSize` (`LIMIT 0`/`LIMIT -1` would otherwise never terminate) so a future #137 config error fails loudly, plus tests for it and for the exact-multiple-of-batch-size and exactly-at-cutoff cases.

**Non-blocking follow-up (belongs with #137):** only `audit_log` has a single-column index on its age column; the other categories' age columns sit as non-prefixes of composite indexes, so a very large first purge scans rather than seeks. Append-mostly tables keep expired rows at low rowids so the `LIMIT` batches find them early, but if large-fleet purges prove slow once #137 drives them on a cadence, a dedicated index per age column is the fix. Noting it here so #137 can evaluate it under a realistic run.

---

PR #420, opened 2026-08-23
feat(enforcement): push date-specific exception overrides to the client + revert (#399)

### LINKED ISSUE #399: Enforcement: push date-specific exception overrides to the client when their window opens (and revert)
## Context

[#142](https://github.com/BenSeymourODB/linux-parental-controls-toolkit/issues/142)
/ PR #398 (ADR 0012) composed date-specific **exceptions** (`allow`/`deny`/`extend`,
own + inherited group) into the effective-policy resolver: they now shape the
per-day `allowedWindows` served by `GET /api/users/:userId/effective` (the
display / `?date=` preview surface). This is deliberately the **resolver /
display** slice.

What it does **not** do — and can't, structurally — is push a one-off override
to the client's `timekpra` allowed-hours. That push is a **static weekly grid**
(`policy/weekly-windows.ts` → `transport/timekpr/allowed-hours.ts`): it resolves
`effectivePolicy` for the seven weekdays of a reference week and keys the result
by ISO weekday. A one-off calendar date is not a weekly-recurring pattern, so
`weekly-windows.ts` deliberately omits exceptions (ADR 0012 §3).

So a "no screen time on 2026-06-30" / "screen-free vacation week" override is
**correct in the dashboard's effective view but not yet enforced on the device.**

## Scope

- A scheduler (croner) that, as a date-specific override's window **opens**,
  pushes the adjusted allowed-hours / daily limit for the affected user(s) to
  their client(s) via the existing SSH + `timekpra` transport, and **reverts**
  to the standing recurring policy when the window closes.
- Queue + replay for offline clients
  ([#84](https://github.com/BenSeymourODB/linux-parental-controls-toolkit/issues/84)),
  exactly like the standing policy push.
- Audit every override push/revert like other commands.
- Decide the representation: a per-day allowed-hours override vs. a
  daily-limit override, and how it composes with the recurring grid already on
  the device (the resolver's per-day `allowedWindows` is the source of truth).
- Tests over the injected SSH executor seam.

## Out of scope

- Per-activity **quota** reduction from `deny` rules (recurring schedules don't
  do this today either) — a separate resolver/enforcement question noted in
  ADR 0012.

## References

- ADR 0012 (`docs/adr/0012-date-specific-override-composition.md`) §3, #142,
  #84, `docs/roadmap.md` → Phase 13.

BenSeymourODB:
## Summary

Delivers the enforcement follow-up ADR 0012 §3 deferred: date-specific
`exceptions` (`allow` / `deny` / `extend`, own + inherited group) now actually
reach the client's `timekpra` allowed-hours **when their window arrives**, and
**revert** to the standing recurring policy when it closes. Until now a
"screen-free vacation week" / "allow games until 9pm tonight" override was
correct in the dashboard's effective view but never enforced on the device.

- **Phase 1 — thread exceptions through the push stack.** Optional, default-off
  seam on `policy/weekly-windows`, `transport/policy-push/{resolve,
  platform-runner,linux-runner,executor}`: the executor's new `includeExceptions`
  loads `gatherUserExceptions` and folds the user's active overrides into the
  allowed-hours grid. Seconds limits are untouched (an exception carries no time
  amount, ADR 0012). The standing push is byte-identical (recurring grid stays
  exception-free, ADR 0012 §3).
- **Phase 2 — the croner scheduler.** A `policy.push.exceptions` kind + a
  `transport/exception-push` scheduler that, each tick, pushes the
  exception-inclusive weekly grid for users with a materially-active override
  (change-detected, 8-day lookback for restart/revert reconciliation) via the
  existing offline queue + audit, and re-resolves at drain time so an expired
  override auto-reverts. Config `PCT_EXCEPTION_PUSH_CRON`, wired in
  `createPolicyPushTransport`.

## Plan

Linked plan: `docs/plans/399-date-override-enforcement-push.md`
Roadmap: `docs/roadmap.md` → Phase 13 (ADR 0012 §3 follow-up)

## Design decisions (grounded in ADR 0012)

- **Representation: per-day allowed-hours override, not a daily-limit override** —
  exceptions move only `allowedWindows`, never seconds (an additive amount is a
  `Grant`). Settles the issue's open representation question.
- **Standing push stays exception-free** so it is the clean recurring baseline /
  revert target; a one-off date can't be a weekly-recurring pattern, so folding
  it into the standing grid would alias onto the next same-weekday. The dedicated
  scheduler pushes the override-inclusive grid and reconciles at least daily
  (default every 15 min), well within the 7-day aliasing horizon (8-day lookback
  guarantees the revert fires once the override's week rolls past).
- **Only `overall`-scoped exceptions have effect today** (activity/group are
  no-ops per ADR 0012) — handled for free: a no-op override yields the standing
  grid, so nothing is pushed.

## License-boundary note

None touched. All changes are pure TypeScript over the policy model + the
existing SSH-subprocess `timekpra` client and the offline queue. No GPL code is
linked in-process; no GPL binary is added to the image (`license-guard`
unaffected).

## No new dependencies

Uses `croner`, `drizzle-orm`, `zod`, and Node built-ins already in the tree.

## Test plan

- [x] Phase 1 — `weekly-windows` override folding (deny / extend-past-a-deny /
  out-of-week / empty == recurring); `resolvePolicyPush` override grid vs.
  unchanged seconds; executor `includeExceptions` on/off, expired-override
  auto-revert, inherited-group override.
- [x] Phase 2 — scheduler tick behaviour (push on override open, revert once
  after the override's week rolls past, first-tick restart reconcile,
  multi-client fan-out, offline → queued, group-inherited, activity-scoped
  no-op, non-retriable isolation, lifecycle) + config default/override/invalid.
- [x] Full gate green: `format:check` / `lint` / `typecheck` clean, **2000**
  unit tests, coverage **98.91%** (≥ 80%).

## Deferred (tracked)

- Admin policy save during an active override transiently clobbers the override
  grid (the standing push is exception-free); the next scheduler tick re-asserts
  it — filed as **#421**.
- Per-activity / group `deny` quota reduction — out of scope per ADR 0012.
- Whole-week deny → full lockout — the existing Phase 8c gap, unchanged (a
  single-day deny within an otherwise-allowed week pushes fine).

Closes #399

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -127,3 +127,11 @@
 # PCT_TIMEKPR_MIRROR_DIR=/data/apt/timekpr    # apt repo lives here
 # PCT_TIMEKPR_MIRROR_PACKAGE=timekpr-next     # or timekpr-next-beta
 # PCT_TIMEKPR_MIRROR_VERSION=0.5.5            # optional: pin a version; newest if unset
+
+# --- Date-specific override enforcement push (optional; #399) ----------------
+# croner cadence at which the dashboard pushes a supervised user's active
+# date-specific overrides (allow/deny/extend for a calendar date or range) to
+# their client's Timekpr-nExT allowed-hours, and reverts to the standing
+# recurring policy once the override lapses. Started only when the live SSH
+# transport exists. Defaults to every 15 minutes.
+# PCT_EXCEPTION_PUSH_CRON=*/15 * * * *
--- a/docs/plans/399-date-override-enforcement-push.md
+++ b/docs/plans/399-date-override-enforcement-push.md
@@ -1,0 +1,154 @@
+# Plan — #399 Push date-specific exception overrides to the client (and revert)
+
+Roadmap: `docs/roadmap.md` → Phase 13 (the enforcement follow-up ADR 0012 §3 defers).
+Issue: [#399](https://github.com/BenSeymourODB/linux-parental-controls-toolkit/issues/399).
+Builds on: #142 / PR #398 (ADR 0012, resolver-side exception composition), #84
+(offline queue + replay), #143 (effective-policy resolver), #140 (weekly
+allowed-hours grid), #232 (platform runner seam).
+
+## The gap
+
+ADR 0012 composed date-specific `exceptions` (`allow` / `deny` / `extend`, own +
+inherited group) into the effective-policy resolver: they now shape the per-day
+`allowedWindows` served by `GET /api/users/:userId/effective` (the display /
+`?date=` preview). ADR 0012 §3 is explicit that this is the **resolver/display**
+slice only, and that **pushing** a date-specific override to the client's
+`timekpra` allowed-hours *when its window arrives* (and reverting after) is a
+separate offline-queue scheduler concern — this issue.
+
+Today the standing push (`transport/policy-push`, `policy/weekly-windows.ts`)
+resolves the **recurring** weekly grid with **no exceptions**, so a
+"screen-free vacation week" / "allow games until 9pm tonight" override is correct
+in the dashboard but never reaches the device.
+
+## Key facts that shape the design (from ADR 0012)
+
+1. **Exceptions move only `allowedWindows`, never seconds.** An additive time
+   amount is a `Grant`, not an exception. So the representation is a **per-day
+   allowed-hours override**, *not* a daily-limit override — the issue's open
+   "allowed-hours vs daily-limit" question is settled by the ADR.
+2. **Only `overall`-scoped exceptions have an observable effect today.**
+   `activity` / `group` exceptions are gathered and composed but resolve to a
+   no-op until per-target deny enforcement lands (out of scope here, as in the
+   ADR). The design handles this for free: a no-op exception produces the same
+   grid as standing, so nothing is pushed.
+3. **Exceptions must stay out of the *recurring* weekly grid.** `timekpra`
+   allowed-hours is a static weekly grid keyed by ISO weekday; a one-off calendar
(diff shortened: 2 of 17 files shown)


BenSeymourODB:
First-pass self-review done (deep read of the full diff against `CLAUDE.md` + ADR 0012, with license-boundary / type-safety / tamper-scope checks — all clean). It surfaced one real blocker, now fixed in `8ca93d9`:

- **Blocker — fail-open on a `deny` override.** The scheduler used change-detection (push only when the desired grid changed). But a standing policy push can clobber the device's allowed-hours out-of-band — it shares the override push's coalesce key and resolves the *exception-free* grid — and a common trigger (any budget/seconds edit, which doesn't change the allowed-hours signature) would silently and permanently drop an active override. The scheduler can't observe that clobber, so it now **re-asserts every tick while an override is materially active**; any clobber is bounded to ≤ one cron interval. Regression test added.

Also addressed from the review:
- First-pass restart sweep now reconciles any user with **any** exception row, so an override that expired during an outage longer than the 8-day steady-state lookback still reverts once on restart (+ test).
- Added the missing `extend`-widens-past-a-standing-deny test in the push resolver.
- Documented `PCT_EXCEPTION_PUSH_CRON` + the scheduler in `docs/server-deployment.md`; softened the "can never outlive" docstring (a Monday-start override's revert coincides with its aliasing Monday, so it can linger up to one cron interval — narrow, documented).

The residual bounded-transient clobber window is tracked in #421. Full gate green: format/lint/typecheck clean, 2002 unit tests, coverage 98.91%.

---

Issue #356, opened 2026-07-02
Client address liveness: detect IP changes on the client (DHCP vs static awareness) and re-announce to the server


BenSeymourODB:
## Context

Even with IPs captured at enrol (see the enrolment-metadata issue), a DHCP client's address can change and silently strand the server's SSH target — the "timeout / stale address" class of the v0.1.0-alpha.5 unreachable incident. The client side already has an authenticated, outbound channel that survives address changes: the per-client bearer token and the `pct-client-bridge` WebSocket (#101), which reconnects client→server regardless of what the server thinks the client's address is.

## What

1. **Re-announce on change.** `pct-client-bridge` watches for address changes (netlink subscription, or a cheap periodic `ip -j addr` diff) and re-announces its current address set to the server over an authenticated endpoint (or as a frame on the event stream once connected). The server updates the client's last-known address(es) + timestamp and, if the admin pinned the SSH target to a now-stale IP, flags the mismatch on the Clients card.
2. **Passive corroboration.** On every event-stream connect the server already knows the socket's remote address — record it as `lastObservedIp` for free, independent of client self-reports. Disagreement between observed and configured target is itself a strong "your SSH target is stale" signal to badge.
3. **DHCP/static awareness.** Best-effort detection on Mint/NetworkManager: `nmcli -t -f ipv4.method connection show <active>` (`auto` = DHCP, `manual` = static). Where detection is inconclusive, fall back to an explicit admin toggle on the client record. Use it to tune UX only (e.g. suggest a DHCP reservation, choose a sensible re-check cadence) — the re-announce mechanism itself works the same either way.
4. **Docs.** Recommend DHCP reservations (or static leases) for supervised machines in `docs/client-install.md` as the low-tech mitigation.

## Notes / boundaries

- This is liveness bookkeeping, not tamper-resistance: a supervised user unplugging Ethernet is out of scope by design (CLAUDE.md → "Tamper resistance is deliberately bounded").
- The re-announce endpoint must be scoped to the client's own bearer token and rate-limited; it may only update *that* client's address fields.
- Server-side auto-retarget (switching the SSH target to the announced IP automatically) should be opt-in; default is to surface the mismatch and let the admin confirm.

## Overlap

- Builds on #101 (bridge) and the enrolment-metadata issue (address fields + SSH target override).
- Complements #350 (outbound upgrade channel spike) — both lean on the client-initiated connection to sidestep server→client reachability problems; neither depends on the other.

[...]

Issue #195, opened 2026-06-19
Usage normalisation: domain / domain_group resolution from web-proxy telemetry


BenSeymourODB:
## What

Resolve `domain`- and `domain_group`-kind `Activity` rows into `UsageSample`
rows, completing the matcher-grammar work begun in #178.

#178 shipped the activity matcher grammar (ADR 0006) and applied it to the two
**window**-resolvable kinds (`app`, `app_group`) in the ActivityWatch
normaliser. It explicitly **deferred** `domain` / `domain_group` because those
kinds match **web requests**, not window events — their telemetry source is the
web proxy (e2guardian, Phase 6 #90) / AdGuard (Phase 7), which does not exist
yet.

## Scope (once web-proxy telemetry lands)

- A normaliser path that takes web-request telemetry (host + timing) and
  resolves each request host against `domain` / `domain_group` activities using
  the **same** `policy/activity-matcher.ts` grammar already built in #178
  (`exact` / `substring` / `glob` / `regex`, case-insensitive, exact-beats-
  pattern then lowest-id precedence). `domain_group` is the host-side analogue of
  `app_group`.
- Map resolved request usage into `usage_samples` rows (the burndown /
  per-activity readers are kind-agnostic, so they need no change).
- Define how request-level usage is intervalised (web requests are point events,
  unlike window durations) — likely a per-host session/coalescing rule; settle
  it in the ADR or a short addendum when the telemetry shape is known.

## Dependencies

- **Blocked by:** the web-proxy telemetry pull (e2guardian, Phase 6 #90 /
  AdGuard Phase 7) — there is no domain-usage source to normalise until then.
- **Builds on:** #178 (the shared matcher grammar `policy/activity-matcher.ts`
  + ADR 0006, which already records the intended `domain*` contract).

Filed as the tracked follow-up for the deferred half of #178.

PR #424, opened 2026-08-23
feat(dns): per-client domain blocklists + active-mode surface (#97)

### LINKED ISSUE #97: Admin UI: per-client domain blocklists with schedule support + active-mode surface
## What

The `/admin` surface for DNS filtering: per-client domain blocklists with schedule support, plus a clear display of which AdGuard **mode** is active so the admin understands where their DNS rules actually land. The AdGuard mode selector is mocked in PR #60 — [`design/admin/integrations.html`](../blob/main/design/admin/integrations.html).

## Scope

- Editor for per-client (and/or per-user) domain blocklists with schedule windows.
- Read the active `PCT_ADGUARD_MODE` (from #95) and show it prominently — including the `external` warning that writes are confined to the `pct:`-prefixed clients/account.
- Writes go through `/api/*` → the AdGuard REST client (#94); no direct UI→AdGuard calls.

## Notes

- Phase 7 "UI: per-client domain blocklists with schedule support; surfaces the active mode" in [`docs/roadmap.md`](../blob/main/docs/roadmap.md).
- Disabled mode should render an explanatory empty state, not a broken editor.

BenSeymourODB:
## Summary

The Phase-7 DNS-filtering admin surface (#97): the `/admin` view for per-client
domain blocklists plus a clear display of the active AdGuard **mode**, backed by
a new write path that composes domain-deny policy into AdGuard rules and pushes
it — REST-only, staying within the license boundary. Delivered end-to-end
(backend + API + frontend) in three phases.

- **`policy/domain-denies.ts`** — extracts the shared always-on `domain`-deny
  resolver (`resolveAlwaysOnDomainDenies` / `isAlwaysOnRule` /
  `domainsForDenyRule`) as a policy-layer concern, and refactors the Phase-6
  e2guardian path to consume it, so the two enforcement mechanisms can never
  disagree about which domains a user always-on denies. Behaviour is identical —
  the e2guardian tests guard the refactor.
- **`transport/adguard/blocklist.ts`** — `buildDnsBlocklistPlan` (per dashboard
  `Client`/device: the union of its users' always-on `domain` denies, mapped to
  one `pct:` AdGuard client keyed on `reportedIps`; devices with denies but no
  reported IP are surfaced as `skipped`, not dropped), `composeUserRules` (a pure
  marker read-modify-write over the global `user_rules` that preserves every
  foreign, hand-written rule), `reconcileManagedClients`, and `applyDnsBlocklist`
  (orchestrator; writes only when the rule list actually changed).
- **API** — `GET /api/dns/blocklist` (preview) + `POST /api/dns/blocklist/apply`
  (admin-only; `409` when DNS is disabled/unconfigured).
- **Frontend** — `DnsFilteringView` + a "DNS filtering" nav item: the active-mode
  banner (mode + health, the `external`-mode confinement warning, a `disabled`
  explanatory empty state), the per-device blocklist table, a "not enforced"
  section for IP-less devices, and an Apply action.
- **ADR 0015** pins the model; **`docs/plans/97-*`** the plan; `docs/server-deployment.md`
  documents the surface + corrects the external-mode rule mechanism.

## Design decisions (ADR 0015)

- **Per-device, not per-user.** A DNS query carries no Linux UID, so AdGuard
  filters per device/IP (per-user web filtering stays e2guardian's job, Phase 6).
  Each device becomes one `pct:<friendlyName ?? hostname>` client keyed on its
  reported IPs; the pushed denies are the union across the device's users.
- **Rule ownership via a marker block** inside AdGuard's global `user_rules`,
  because `set_rules` replaces the whole list. Apply strips the previous marked
  block, keeps all foreign rules, and appends a freshly-composed, sorted block —
  idempotent.
- **v1 = always-on denies only.** `user_rules` have no native scheduling, so
  windowed/date-scoped DNS denies (which need a rule-swap scheduler) are deferred
  to **#425**, mirroring how e2guardian shipped always-on (#90) before windows (#216).

## Plan

Linked plan: `docs/plans/97-adguard-dns-blocklist-admin.md`
ADR: `docs/adr/0015-adguard-dns-filtering-model.md`
Roadmap: `docs/roadmap.md` → Phase 7 (DNS filtering).

## License-boundary note

AdGuard Home is driven over its REST API only (`CLAUDE.md` → "License
boundaries" rule 4). No AdGuard code is linked/imported/vendored; no GPL binary
added to the image (`license-guard` unaffected). Pure TypeScript + zod + Svelte.

## No new dependencies

Uses `drizzle-orm`, `zod`, the existing AdGuard REST client, and Svelte 5.

## Test plan

- [x] **Backend:** `policy/domain-denies` (always-on collection; group/inherited
  denies; recurring/date-scoped + non-deny excluded) and
  `transport/adguard/blocklist` (plan build incl. skip/merge/union; marker RMW
  incl. idempotence + malformed-block tolerance; reconcile add/update/delete;
  end-to-end apply incl. no-op re-apply).
- [x] **API:** `GET /api/dns/blocklist` (401; disabled → not-applyable preview;
  external healthy → applyable) + `POST /api/dns/blocklist/apply` (401; `409`
  disabled; healthy external reconcile + push, foreign rules preserved).
- [x] **Frontend:** `DnsFilteringView` — external banner + confinement warning +
  table; disabled empty state (no editor); disabled/reasoned Apply; skipped
  devices; apply + summary + re-read; load-error surface.
- [x] Server gate green: `format:check` / `lint` / `typecheck` clean; **2009**
  unit tests; coverage **98.9%** (≥ 80%; new modules at 100% lines).
- [x] Frontend gate green: `svelte-check` 0/0; **381** tests (6 new); static build OK.

## Deferred (tracked)

- **Scheduled / time-windowed / date-scoped DNS denies** (rule-swap scheduler) → **#425** (filed).
- `domain_group` expansion + matcher-kind-aware translation → **#178 / #195**.
- Dynamic-IP reconciliation → **#356**.
- Auditing DNS pushes — `audit_log` is transport-command-shaped (client target +
  argv); a REST rule push has neither, so left out of v1 (same reason the
  retention purge used a dedicated ledger).

Closes #97


BenSeymourODB:
Ran a first-pass adversarial self-review (license boundaries, the `user_rules` read-modify-write, reconcile, the e2guardian refactor, types). License boundary is clean — AdGuard is only ever reached over REST, no GPL linkage, no image change. Three findings were worth acting on and are fixed in `e949621`:

1. **Client filtering inheritance (correctness — the one worth calling out).** New `pct:` clients were created with just `{name, ids}`. AdGuard's add-client API defaults an omitted `use_global_settings` to `false` (Go zero value), which would disable filtering for that client and make the pushed `$client=` rules a **silent no-op** — clients created, rules written, and nothing blocked. Managed clients are now created/kept with `use_global_settings: true` explicitly (and a mismatched existing one self-heals on the next apply). ADR 0015 updated to record this. Worth a sanity check against a real instance when you next have one up, but setting it explicitly makes the code correct regardless of AdGuard's default.
2. **Apply health-gate.** `applyable` (and the apply guard) now require `health === "ok"`, not just a wired client, so an unreachable/unhealthy external instance surfaces as not-applyable (`409`) instead of failing mid-push with an opaque `500`; a transport error during apply maps to a clean `502`.
3. **Merge ordering.** `buildDnsBlocklistPlan` now merges same-name devices *before* the enforce-vs-skip decision, so an IP-less device can't strand the domains of a same-named device that does have an address (and a name never appears in both `clients` and `skipped`).

Also added HTTP coverage for the delete-stale / foreign-survives path and the unreachable → `409` path. Full gate green (2014 unit tests, coverage 98.9%; frontend 381, build OK); all CI checks green on the previous head.

---

